PR #7, opened 2024-05-15
add checking of return status on fstat calls

ncroxon:
There are a few places we don't check the return status when calling fstat for success. Clean up the calls by adding a check before continuing.

V2: clean up warnings/errors from checkpatch.pl

### DIFF
--- a/Assemble.c
+++ b/Assemble.c
@@ -652,7 +652,9 @@ static int load_devices(struct devs *devices, char *devmap,
 			/* prepare useful information in info structures */
 			struct stat stb2;
 			int err;
-			fstat(mdfd, &stb2);
+
+			if (fstat(mdfd, &stb2) != 0)
+				goto error;
 
 			if (c->update == UOPT_UUID && !ident->uuid_set)
 				random_uuid((__u8 *)ident->uuid);
@@ -675,13 +677,10 @@ static int load_devices(struct devs *devices, char *devmap,
 				       devname);
 				if (dfd >= 0)
 					close(dfd);
-				close(mdfd);
-				free(devices);
-				free(devmap);
 				tst->ss->free_super(tst);
 				free(tst);
 				*stp = st;
-				return -1;
+				goto error;
 			}
 			tst->ss->getinfo_super(tst, content, devmap + devcnt * content->array.raid_disks);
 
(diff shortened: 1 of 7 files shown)


Assemble.c
```diff
@@ -652,8 +652,9 @@ static int load_devices(struct devs *devices, char *devmap,
 			/* prepare useful information in info structures */
 			struct stat stb2;
 			int err;
-			fstat(mdfd, &stb2);
-
+			if (fstat(mdfd, &stb2) != 0) {
+				goto error;
+			}
```
mtkaczyk, line 657: You don't need braces in this case, there is one instruction inside if but it is not an issue, just FYI,

Dump.c
```diff
@@ -110,9 +111,17 @@ int Dump_metadata(char *dev, char *dir, struct context *c,
 		free(fname);
 		return 1;
 	}
-	if (c->verbose >= 0)
+	if (c->verbose >= 0) {
```
mtkaczyk, line 114: Oh, bracases here totally change the flow.

Previously, only `printf()` was executed verbose >=0. Now, the fsttat is executed. I think that it should be:
``` suggestion
	if (c->verbose >= 0)
		printf("%s saved as %s.\n", dev, fname);
	close(fl);
	ret = fstat(fd, &dstb);
	close(fd);

	if (ret) {
		unlink(fname);
		free(fname);
		return 1;
	}
```

mtkaczyk (changes requested):
Hey @ncroxon,
There is one small issue. Please fix it and we are ready to go :)

ncroxon:
I pushed a new version (V3) to the branch.

mtkaczyk:
```
[PATCH V3] add checking of return status on fstat calls
There are a few places we don't check the return status when
calling fstat for success. Clean up the calls by adding a
check before continuing.

V3: fix Dump.c braces to execute verbose properly.
    clean up braces with only one instruction inside.
V2: clean up warnings/errors from checkpatch.pl

Signed-off-by: Nigel Croxon <ncroxon@redhat.com>
```

I will cherry-pick this commit so please remove [PATCH V3] from title and v2 and v3 changes from description.
We have to use comment to note changes from now.. thanks!

PR #8, opened 2024-05-15
Makefile: Do not call gcc directly

gwendalcr:
When mdadm is compiled with clang, direct gcc will fail. Make sure to use $(CXX) variable instead.


mtkaczyk:
Hey @gwendalcr,
Could you please add to commit evidence that those commands work with clang. I'm not against it but since mdadm was limited to gcc only I need evidence that the same way we can check clang.

 `$(shell $(CXX) $(CXXFLAGS) $(LDFLAGS) -Q --help=warnings 2>&1 | grep "format-overflow" | wc -l)`
 
 Calls you changed were used to determine availability of particular gcc flag. It was not a compilation setting so `$(CXXFLAGS) $(LDFLAGS)` are not needed.

Also, below `$(CC)` is is used not `$(CXX)` - CXX is probably c++ compiler.  If that is wrong you have to update more files.

For example:

``` C
mdadm : $(OBJS) | check_rundir
	$(CC) $(CFLAGS) $(LDFLAGS) -o mdadm $(OBJS) $(LDLIBS)
```

From my experience, it will not be honored if it is not tested so I would like ask you to propose clang check action. We experienced this problem in different application, so I can support you, please look:
https://github.com/intel/ledmon/blob/main/.github/workflows/review.yml#L51

Thanks,
Mariusz

mtkaczyk (changes requested):
See comment above.

gwendalcr:
Without this patch, compiling with clang fails: in chromeos/gentoo, trying to emerge mdadm-4.3:
```
make -j36 PKG_CONFIG=/build/brya/build/bin/pkg-config CC=x86_64-cros-linux-gnu-clang CWFLAGS=-Wall CXFLAGS= -Os -pipe  -march=alderlake -g -fno-exceptions -fno-unwind-tables -fno-asynchronous-unwind-tables -ffunction-sections -fdata-sections LDFLAGS=-Wl,-O2 -Wl,--as-needed  -Wl,--gc-sections -Wl,--icf=all UDEVDIR=/lib/udev SYSTEMD_DIR=/lib/systemd/system COROSYNC=-DNO_COROSYNC DLM=-DNO_DLM STRIP= all
        -sh -c gcc -Q --help=warnings 2>&1 | grep "stringop-overflow" | wc -l
...
 *   Bad gcc [-Q --help=warnings] invocation
```

Clang does not have `warnings` option in help, therefore, the command always fails and no extra flags are addded:
```
x86_64-cros-linux-gnu-clang -Wall -Os -pipe  -march=alderlake -g -fno-exceptions -fno-unwind-tables -fno-asynchronous-unwind-tables -ffunction-sections -fdata-sections -Q --help=warnings 
clang: error: unknown argument: '--help=warnings'
clang: error: no input files
```
However, when `-print-diagnostic-options` is used, all the flags are available.
Note that gentoo ignore the`CWFLAGS` logic completely, since it defines its value [beforehand](https://gitweb.gentoo.org/repo/gentoo.git/tree/sys-fs/mdadm/mdadm-4.3.ebuild#n51).

mtkaczyk:
```
When mdadm is compiled with clang, direct gcc will fail.
Make sure to use $(CC) variable instead.

Note that Clang does not support `--help=warnings` --
`-print-diagnostic-options` should be used instead.
So with Clang, the compilation will go through, but the
extra warning flags will never be added.

Signed-off-by: Gwendal Grignou <gwendal@chromium.org>
```

Please do not use markdown syntax in commit message.

Please fix: --`-print-diagnostic-options`.

LGTM, please update commit message and I will merge it :)

mtkaczyk:
Please fix typo: 
``` text
--
-print-diagnostic-options.
```

There are three '-'.

mtkaczyk (approved):

mtkaczyk:
Hey @gwendalcr,
According to policy we follow, I have to keep history linear. Github tells me that I will merge 2 commits so I have to ask you to reset the head to current main and cherry-pick the commit on top of it.
There is merge commit I have to take with your patch.

PR #11, opened 2024-05-17
super-intel fix dead code

ncroxon:
Execution cannot reach this statement: "while (devlist) { dv = de...". Local variable "err" is assigned only once, to a constant value, making it effectively constant throughout its scope.

### DIFF
--- a/.checkpatch.conf
+++ b/.checkpatch.conf
@@ -1,0 +1,16 @@
+--no-tree
+--show-types
+--exclude .github
+--exclude clustermd_tests
+--exclude documentation
+--exclude misc
+--exclude systemd
+--exclude tests
+--exclude *.md
+--exclude *.4
+--exclude *.in
+--exclude *.8
+--exclude test
+--exclude udev-*
+--ignore FILE_PATH_CHANGES
+--ignore EMAIL_SUBJECT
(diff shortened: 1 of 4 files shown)


super-intel.c
```diff
@@ -7080,7 +7080,17 @@ get_devices(const char *hba_path)
 		}
 
 		dv = xcalloc(1, sizeof(*dv));
+		if (dv == NULL) {
```
mtkaczyk, line 7083: please see xcalloc implementation and you will realize that the code won't be executed because xcalloc will abort the operation.

If `err` is not used then I think that the error path should be removed.

super-intel.c
```diff
@@ -7080,7 +7080,17 @@ get_devices(const char *hba_path)
...
+			fprintf(stderr, "malloc failed\n");
+			err = 1;
+			break;
+		}
 		dv->devname = xstrdup(buf);
+		if (dv->devname == NULL) {
```
mtkaczyk, line 7089: same as above:

``` C
char *xstrdup(const char *str)
{
	char *rv = strdup(str);
	char *msg;
	int n;
	if (rv)
		return rv;
	msg =  ": memory allocation failure - aborting\n";
	n = write(2, Name, strlen(Name));
	n += write(2, msg, strlen(msg));
	exit(4+!!n);
}
```

mtkaczyk (changes requested):
Hi @ncroxon,
probably you have to remove the error path totally.

Mariusz

Issue #12, opened 2024-05-20
Build fails with musl libc: implicit declaration of function 'basename' 


ncopa:
Building source from git fails with:

```
Monitor.c:257:34: error: implicit declaration of function 'basename' [-Werror=implicit-function-declaration]
  257 |                                  basename(mdlist->devname));
      |                                  ^~~~~~~~
In file included from Monitor.c:25:
mdadm.h:115:20: error: format '%s' expects argument of type 'char *', but argument 4 has type 'int' [-Werror=format=]
  115 | #define DEV_MD_DIR "/dev/md/"
      |                    ^~~~~~~~~~
Monitor.c:256:81: note: in expansion of macro 'DEV_MD_DIR'
  256 |                         snprintf(st->devname, MD_NAME_MAX + sizeof(DEV_MD_DIR), DEV_MD_DIR "%s",
      |                                                                                 ^~~~~~~~~~
Monitor.c:256:94: note: format string is defined here
  256 |                    snprintf(st->devname, MD_NAME_MAX + sizeof(DEV_MD_DIR), DEV_MD_DIR "%s",
      |                                                                                        ~^
      |                                                                                         |
      |                                                                                         char *
      |                                                                                        %d
Monitor.c: In function 'check_one_sharer':
Monitor.c:492:21: error: passing argument 1 of 'strncmp' makes pointer from integer without a cast [-Werror=int-conversion]
  492 |         if (strncmp(basename(comm), Name, strlen(Name)) == 0) {
      |                     ^~~~~~~~~~~~~~
      |                     |
      |                     int
In file included from /usr/include/fortify/string.h:23,
                 from mdadm.h:48:
/usr/include/string.h:40:14: note: expected 'const char *' but argument is of type 'int'
   40 | int strncmp (const char *, const char *, size_t);
      |              ^~~~~~~~~~~~
cc1: all warnings being treated as errors
make: *** [Makefile:211: Monitor.o] Error 1
```

In POSIX `basename` is defined in `libgen.h`, but there is also a [GNU implementation](https://man7.org/linux/man-pages/man3/basename.3.html#VERSIONS). I am not sure which variant `mdadm` intended to use?

Related downstream issue and merge request:
- https://gitlab.alpinelinux.org/alpine/aports/-/issues/16122#note_406272
- https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/63905

ncopa:
Looks like `mdadm` expects GNU basename behavior.

[...]

PR #14, opened 2024-05-21
mdadm.h: provide basename if GLIBC is not avialable

mtkaczyk:
If GNU basename is not avilable, define it. It is safer to use that rather than include libgen.h with XPG basename() definition.

Fixes:#12


mtkaczyk:
Hello @kraj could you please test if this fixes the issue for you?

You reported patch and never answered review on ML:
https://lore.kernel.org/linux-raid/20240327121023.000017e9@linux.intel.com/

Thanks!

ktanska (approved):

PR #17, opened 2024-05-22
Makefile: fix make -s detection

nbuwe:
Only check the first word of MAKEFLAGS for 's', that's where all the single letter options are collected.

MAKEFLAGS contains _all_ make flags, so if any command line argument contains a letter 's', the silent test will be false positive.  Think e.g. make 'DESTDIR=.../aports/main/mdadm/pkg/mdadm' install

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -157,7 +157,7 @@ ifndef UDEVDIR
  UDEVDIR = /lib/udev
 endif
 
-ifeq (,$(findstring s,$(MAKEFLAGS)))
+ifeq (,$(findstring s,$(firstword -$(MAKEFLAGS))))
 	ECHO=echo
 else
 	ECHO=:


mtkaczyk:
Hello @nbuwe,
Thanks for fix!

What about:
```make
ifeq (,$(findstring "-s ",$(MAKEFLAGS)))
```
Limiting check to first word is something that could cause headache for maintainers.

nbuwe:
That's not how `MAKEFLAGS` works.  make collects all the options that have single-letter variant and put them all together as a single word _without_ the leading dash as the first word of `MAKEFLAGS`.  So something like `make -s --keep-going` will result is `ks` in `MAKEFLAGS`.  If there are no short options, but there are long-only options or variable assignments, they will be in `MAKEFLAGS` after a leading space (with an empty first word).  That's why the suggested fix prepends `-`, to make sure the first word exists even if it's originally empty.

In legalese: https://www.gnu.org/software/make/manual/make.html#Options_002fRecursion

> The value of MAKEFLAGS is a possibly empty group of characters representing single-letter options that take no argument, followed by a space and any options that take arguments or which have long option names. If an option has both single-letter and long options, the single-letter option is always preferred. If there are no single-letter options on the command line, then the value of MAKEFLAGS starts with a space.

mtkaczyk:
On top of my Makefile I added:
```make
$(info flags: $(MAKEFLAGS))
```
and I see:
```bash
# make clean -s
flags: s                                                                                       │ 
```
Thanks for clarification. I proposed it because it somehow worked for me but I didn't dig deep enough to understand what is going on.

With the same testing approach I also see:
```bash
# make clean 'DESTDIR=.'
flags:
rm -f mdadm mdmon mdadm.o config.o policy.o mdstat.o  ReadMe.o uuid.o util.o maps.o lib.o udev.o Manage.o Assemble.o Build.o Create.o Detail.o Examine.o Grow.o Monitor.o dlink.o Kill.o Query.o Incremental.o Dump.o mdopen.o super0.o super1.o super-ddf.o super-intel.o bitmap.o super-mbr.o super-gpt.o restripe.o sysfs.o sha1.o mapfile.o crc32.o sg_io.o msg.o xmalloc.o platform-intel.o probe_roms.o crc32c.o drive_encryption.o mdmon.o monitor.o managemon.o uuid.o util.o maps.o mdstat.o sysfs.o config.o mapfile.o mdopen.o policy.o lib.o udev.o Kill.o sg_io.o dlink.o ReadMe.o super-intel.o super-mbr.o super-gpt.o super-ddf.o sha1.o crc32.o msg.o bitmap.o xmalloc.o platform-intel.o probe_roms.o crc32c.o drive_encryption.o pwgr.o core *.man \
mdadm.tcc mdadm.uclibc mdadm.static *.orig *.porig *.rej *.alt \
.merge_file_* mdadm.Os mdadm.O2 mdmon.O2 swap_super init.cpio.gz \
mdadm.uclibc.static test_stripe raid6check raid6check.o mdmon mdadm.8
rm -rf cov-int
```

I'm not sure if this testing method in convenient but I don't see the problem you referenced in description, i.e. `$(MAKEFLAGS)` are empty in this case.

nbuwe:
The `DESTDIR` example in the commit message is the actual reproducer that triggered me to look into this.

```
$ cat flags.mk
$(info flags: $(MAKEFLAGS))
default:
	@echo MAKEFLAGS=\"$(MAKEFLAGS)\"
$ make -f flags.mk DESTDIR=.
flags: 
MAKEFLAGS=" -- DESTDIR=."
```

mtkaczyk:
Thanks. Got it.
Merged!

PR #18, opened 2024-05-22
mdadm: super-intel remove dead code

ncroxon:
Execution cannot reach this statement: "while (devlist) { dv = de...". Local variable "err" is assigned only once, to a constant value, making it effectively constant throughout its scope. Remove dead code.

### DIFF
--- a/super-intel.c
+++ b/super-intel.c
@@ -7046,7 +7046,6 @@ get_devices(const char *hba_path)
 	struct md_list *dv;
 	struct dirent *ent;
 	DIR *dir;
-	int err = 0;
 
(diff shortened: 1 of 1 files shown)


peluse (commented):
Never looked at this code before but I wonder if the error condition a few lines up was intended to set err?  Otherwise whatever that, what appears to be a single device error, is really just a warning.  Maybe some more insight into that error case would be good before removing the code that basically frees the mem and returns NULL?

ncroxon:
The old code set err. There were modifications to clean up the flow and they never removed the stale code.
I have closed this PR.. and have opened it again, under a new number.

PR #23, opened 2024-06-03
test: pass flags to services

mku514k:
Commit 4c12714d1ca0 ("test: run tests on system level mdadm") removed MDADM_NO_SYSTEMCTL flag from test suite. This causes imsm tests to fail as mdadm no longer triggers mdmon and flags exists only within session.

Use systemd set/unset-environment to pass necessary flags.

Introduce colors to grab users attention to warnings and key messages.

Make test suite setup systemd environment.
Add setup/clean_systemd_env() functions.
Warn user about altering systemd environment.

Add colors to success/fail messages and warnings.

### DIFF
--- a/test
+++ b/test
@@ -127,7 +127,7 @@ do_test() {
 				dmesg | grep -iq "error\|call trace\|segfault" | grep -v "systemd" &&
 					die "dmesg prints errors when testing $_basename!"
 			fi
-			echo "succeeded"
+			succeed "succeeded\n"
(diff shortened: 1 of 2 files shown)


mtkaczyk:
Added to mainline, thanks!

Comment in case of someone will complain about colors:
I'm not against colors for better message filtering, it is ignored if not supported. That is fine for me.
If there will be issues with this functionality- we can remove it.

PR #24, opened 2024-06-04
mdadm: Fix socket connection failure when mdmon runs in foreground mode.

shmsingh:
While creating an IMSM RAID, mdadm will wait for the mdmon main process to finish if mdmon runs in forking mode. This is because with "Type=forking" in the mdmon service unit file, "systemctl start service" will block until the main process of mdmon exits. At that moment, mdmon has already created the socket, so the subsequent socket connect from mdadm will succeed.

However, when mdmon runs in foreground mode (without "Type=forking" in the service unit file), "systemctl start service" will return once the mdmon process starts. This causes mdadm and mdmon to run in parallel, which may lead to a socket connection failure since mdmon has not yet initialized the socket when mdadm tries to connect. If the next instruction/command is to access this device and try to write to it, a permission error will occur since mdmon has not yet set the array to RW mode.

### DIFF
--- a/msg.c
+++ b/msg.c
@@ -151,6 +151,7 @@ int connect_monitor(char *devname)
 	struct sockaddr_un addr;
 	int pos;
 	char *c;
+	int rv, retry_count = 0;
 
 	pos = sprintf(path, "%s/", MDMON_DIR);
 	if (is_subarray(devname)) {
@@ -170,7 +171,24 @@ int connect_monitor(char *devname)
 
 	addr.sun_family = PF_LOCAL;
 	strcpy(addr.sun_path, path);
-	if (connect(sfd, (struct sockaddr*)&addr, sizeof(addr)) < 0) {
+
+	/* In foreground mode, when mdadm is trying to connect to control
+	 * socket it is possible that the mdmon has not created it yet.
+	 * Give some time to mdmon to create socket.
+	 */
+	for (retry_count = 0; retry_count < 10; retry_count++) {
+		rv = connect(sfd, (struct sockaddr*)&addr, sizeof(addr));
+
+		if (rv < 0) {
+			sleep_for(0, MSEC_TO_NSEC(200), true);
+			continue;
+		}
+		break;
+	}
+
+	if (rv < 0) {
+		pr_err("Failed to connect to control socket. (%s!!)\n",
+				strerror(errno));
 		close(sfd);
 		return -1;
 	}


mtkaczyk:
Hey @shmsingh,
Isn't it fixed by https://github.com/md-raid-utilities/mdadm/commit/aa1cc5815d2b14a8b47add18cfaa8264e19c10ce?

shmsingh:
Hi @mtkaczyk,

In my case the race is happening in ping_monitor(). mdadm ping_monitor() will call connect_monitor() which will try to connect to the control socket which mdmon will create. But in foreground mode, since mdamd is not waiting for mdmon to finish creating the socket,  the connect() will fail. 
The fix [https://github.com/md-raid-utilities/mdadm/commit/aa1cc5815d2b14a8b47add18cfaa8264e19c10ce](url) is not addressing the issue. In mdmon we will create pid file followed by control socket file. And we are accessing control socket in ping_monitor in mdadm while in this fix we are calling wait_for_mdmon() after trying to access to the control socket. So the issue I am seeing still persist.

msg.c
```diff
@@ -170,11 +171,26 @@ int connect_monitor(char *devname)
...
+	while (1) {
+		rv = connect(sfd, (struct sockaddr*)&addr, sizeof(addr));
+		if (rv < 0 && retry_count >= 10) {
+			pr_err("Failed to connect to control socket. (%s!!)\n",
+					strerror(errno));
+			close(sfd);
+			return -1;
+		} else if (rv < 0) {
+			retry_count++;
+			sleep(1);
```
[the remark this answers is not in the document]
mtkaczyk: I think one second it too big, what about 200 milliseconds?

This path is to critical, I understand the issue but we cannot accept 10 second wait on metadata write path.
I see that new connection is established every time on `flush_metadata_updates()`.

If you need to initialize the connection to see if mdmon is really working, then you should create new function for this particular usecase.

Please also note that udev may kill the process that is working to long. It is not your usecase but the code you will create will be executed from udev context. For that reason I recommend wait up to 2 seconds.

mtkaczyk (changes requested):

mtkaczyk:
Hey @shmsingh,
Thanks. Please take a look into my comments. IMO it is too risky that is why I requested some changes.

msg.c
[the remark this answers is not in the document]
shmsingh: Thanks for your suggestions. I am working on the changes and I will run the test script overnight with new changes and see the status. If I didn't see the issue than I think changing the waiting period to 2 seconds is good.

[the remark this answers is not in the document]
shmsingh: Hi @mtkaczyk, I ran the test case overnight with delay cap set to 2 seconds and didn't observe any failure. I have updated the patch. Can you please review it. 
Thanks

msg.c
```diff
@@ -170,11 +174,26 @@ int connect_monitor(char *devname)
...
-		return -1;
-	}
 
+	/* In foreground mode, when mdadm is trying to connect to control
+	 * socket it is possible that the mdmon has not created it yet.
+	 * Give some time to mdmon to create socket.
+	 */
+	while (1) {
+		rv = connect(sfd, (struct sockaddr*)&addr, sizeof(addr));
+		if (rv < 0 && retry_count >= 10) {
```
mtkaczyk, line 184: In fact we are waiting 2,2 second (we are waiting 11 times) :) because of `retry_count >= 10`, it should be `retry_count > 10`

msg.c
```diff
@@ -170,11 +174,26 @@ int connect_monitor(char *devname)
...
+			pr_err("Failed to connect to control socket. (%s!!)\n",
+					strerror(errno));
+			close(sfd);
+			return -1;
+		} else if (rv < 0) {
+			retry_count++;
+			nanosleep(&delay, &delay);
```
mtkaczyk, line 191: we have `sleep_for()` you can use.

msg.c
```diff
@@ -170,11 +174,26 @@ int connect_monitor(char *devname)
...
+			continue;
+		} else {
+			break;
+		}
+	}
```
mtkaczyk, line 196: Please add empty line after the loop end.

msg.c
mtkaczyk: I propose reworking it to `for` better readability (`for` suggest that we are iterating over something), what do you think about it?
```c
int rv = 0;

for (retry_count = 0; retry_count < 10; retry_count ++) {
    rv = connect(sfd, (struct sockaddr*)&addr, sizeof(addr));

    if (rv < 0) {
           // I didn't check if I put correct value
           sleep_for(0, 200, false);
           continue;
    }
    break;
}

if (rv) {
     pr_err("Failed to connect to control socket. (%s!!)\n", strerror(errno));
     close(sfd);
     return -1;
}
```

shmsingh: looks good. I will update the patch.

msg.c
shmsingh: I wasn't sure about the 3rd parameter value, that's why I went with using nanosleep. You have used 3rd parameter (wake_after_interrupt) as false in this case. I believe that is the case in the scenario.

mtkaczyk: Hmm, I looked into code and this parameter is set to true everywhere, also setting it to true will produce same result as as change you proposed (and tested). Please set it to true then, not false.

shmsingh:
@mtkaczyk  I have updated the patch. Can you please review it. Thanks

mtkaczyk:
Assuming that the change is safe because it extends current functionality (we do not observe fails with one write only), I'm merging this.

PR #27, opened 2024-06-18
Makefile: Do not call gcc directly (v2)

mtkaczyk:
When mdadm is compiled with clang, direct gcc will fail. Make sure to use $(CC) variable instead.

Note that Clang does not support --help=warnings,
--print-diagnostic-options should be used instead. So with Clang, the compilation will go through, but the extra warning flags will never be added.

New PR to fix merge.


mtkaczyk:
Mention: #8

PR #30, opened 2024-06-20
Revert "mdadm: Fix socket connection failure when mdmon runs in foreground mode."

mtkaczyk:
Reverts md-raid-utilities/mdadm#24

@shmsingh I was over optimistic and this change is not ready. The first is that I suggested you to use `for` and I didn't notice that you use `errno` in error message so I gives us following error:
 ``` mdadm: Failed to connect to control socket. (Success!!)  ```
If timeout is reached, errno contains last `usleep()` result.

We can fix that. Not a big deal. There are more critical problem:
It looks like that `ping_monitor()` is called multiple times not as a hard requirement for monitor to be. As a result, the console is frozen for 2 seconds on various commands. This is not acceptable.
For example:
- creation of volume (Looks simple to fix but it is not, `need_mdmon` logic is not complete)
- stopping raids (Manage_stop)
- reshape flows (Grow.c file)
- adding (Manage.c file)
- Incremental (Incremental.c)

Please see how often `ping_monitor()` is used.

Even I disagree that we should call `ping_monitor()` with particularly to chance of success and no error check- well, this is not what code does. Your change is good in the scenarios where `ping_monitor()` is used correctly but overall code is not ready for it. For that reason, I need to revert it.

If you don't want to fix `ping_monitor()` calls case by case, I would like to propose you creation of `connect_monitor_blocking()` or ping_monitor_blocking() and use that to handle cases where it is obviously required to.

### DIFF
--- a/msg.c
+++ b/msg.c
@@ -151,7 +151,6 @@ int connect_monitor(char *devname)
 	struct sockaddr_un addr;
 	int pos;
 	char *c;
-	int rv, retry_count = 0;
 
 	pos = sprintf(path, "%s/", MDMON_DIR);
 	if (is_subarray(devname)) {
@@ -171,24 +170,7 @@ int connect_monitor(char *devname)
 
 	addr.sun_family = PF_LOCAL;
 	strcpy(addr.sun_path, path);
-
-	/* In foreground mode, when mdadm is trying to connect to control
-	 * socket it is possible that the mdmon has not created it yet.
-	 * Give some time to mdmon to create socket.
-	 */
-	for (retry_count = 0; retry_count < 10; retry_count++) {
-		rv = connect(sfd, (struct sockaddr*)&addr, sizeof(addr));
-
-		if (rv < 0) {
-			sleep_for(0, MSEC_TO_NSEC(200), true);
-			continue;
-		}
-		break;
-	}
-
-	if (rv < 0) {
-		pr_err("Failed to connect to control socket. (%s!!)\n",
-				strerror(errno));
+	if (connect(sfd, (struct sockaddr *)&addr, sizeof(addr)) < 0) {
 		close(sfd);
 		return -1;
 	}


shmsingh:
Hi @mtkaczyk,
Thanks for the detailed explanation. I understood your point. I did some research and came up with few solutions.

- **Revert the foreground patch**: This is a straight forward solution, and we are confident that it does not introduce any known issues since "Type=forking" is being used for almost a decade now.

- **Implement Mariusz's suggestion**: Instead of incorporating this fix into the ping monitor, we could add a new function, connect_monitor_blocking(), and use it on a case-by-case basis. Similar to [aa1cc58](https://github.com/md-raid-utilities/mdadm/commit/aa1cc5815d2b14a8b47add18cfaa8264e19c10ce).

- **Use the fix [aa1cc58](https://github.com/md-raid-utilities/mdadm/commit/aa1cc5815d2b14a8b47add18cfaa8264e19c10ce):** Incorporate a wait_for_mdmon check before calling to ping_monitor in the Create function. 

Please let me know your thoughts on above suggestions.

Thanks

mtkaczyk:
Hey @shmsingh 
I already reverted it. I don't want to keep it because while discussion there is probability that some OSes will take it  (this is really harmful for users). Believe me, mdadm availability is really worldwide.

- Revert the foreground patch.
This is the simplest way with immediate effect. I didn't yet reached the issue described by you but sooner or later I will need to handle it. This is kind of chickening out (we don't want to deal with real root cause, we should wait for mdmon where it is absolutely necessary). This is not complete fix as parent is closed before control socket is initialized.

- Implement Mariusz's suggestion
This is the right way to do it. Each case we have to wait, we are waiting for socket to be up. that is what I would expect.
It requires to care if mdmon is expected. We need to have logic which is able to determine if mdmon is needed (it can be based on raid level of arrays upon containers). Eventually, we can record such information in /var/run/mdadm.map but probably it is not best choice.

- Use the fix [aa1cc58](https://github.com/md-raid-utilities/mdadm/commit/aa1cc5815d2b14a8b47add18cfaa8264e19c10ce)
This is almost same as option number 2 but the difference is that we are checking for pid not for socket:

```c
	/* Ok, this is close enough.  We can say goodbye to our parent now.
	 */
	if (victim > 0)
		remove_pidfile(devnm);
	if (make_pidfile(devnm) < 0) { #pidfile init
		exit(3);
	}
	container->sock = make_control_sock(devnm); #socket init
```
There is small delay so it gives us race possibility, same as in case 1. But we have to add mdmon determination as in case 2.

Issue #22, opened 2024-05-23
Dead code in super1.c


ncroxon:
One line of dead code in super1.c `space = UINT16_MAX;`.

Event cond_at_least: | Condition "space < 16", taking false branch. Now the value of "space" is at least 16.

1448 			`if (space < (PPL_HEADER_SIZE >> 9) + 8) {`
1449 				`pr_err("Not enough space to add ppl\n");`
1450 				`return -2;`
1451 			`}`

Event cond_between: | Condition "space >= 2048", taking false branch. Now the value of "space" is between 16 and 2047.

1453 			`if (space >= (MULTIPLE_PPL_AREA_SIZE_SUPER1 >> 9)) {`
1454 				`space = (MULTIPLE_PPL_AREA_SIZE_SUPER1 >> 9);`
1455 			`} else {`
1456 				`int optimal_space = choose_ppl_space(`
1457 							`__le32_to_cpu(sb->chunksize));`

Event cond_at_most:	Condition "space > optimal_space", taking true branch. Now the value of "optimal_space" is at most 2046.

1458 				`if (space > optimal_space)`

Event assignment:	Assigning: "space" = "optimal_space".

1459 					`space = optimal_space;`

Event at_most:	At condition "space > 65535", the value of "space" must be at most 2047.
Event dead_error_condition:	The condition "space > 65535" cannot be true.

1460 				`if (space > UINT16_MAX)`

Event dead_error_line:	Execution cannot reach this statement: "space = 65535;".

1461 					`space = UINT16_MAX;`

mtkaczyk:
Hi @ncroxon 
Are you going to work on it?

PR #33, opened 2024-06-24
mdadm: Fix socket connection failure when mdmon runs in foreground mode.

shmsingh:
While creating an IMSM RAID, mdadm will wait for the mdmon main process to finish if mdmon runs in forking mode. This is because with "Type=forking" in the mdmon service unit file, "systemctl start service" will block until the main process of mdmon exits. At that moment, mdmon has already created the socket, so the subsequent socket connect from mdadm will succeed.

However, when mdmon runs in foreground mode (without "Type=forking" in the service unit file), "systemctl start service" will return once the mdmon process starts. This causes mdadm and mdmon to run in parallel, which may lead to a socket connection failure since mdmon has not yet initialized the socket when mdadm tries to connect. If the next instruction/command is to access this device and try to  write to it, a permission error will occur since mdmon has not yet set the array to RW mode.


util.c
```diff
@@ -1932,6 +1932,55 @@ int mdmon_running(const char *devnm)
...
 
+/* wait_for_mdmon_control_socket() - Waits for mdmon control socket
+ * to be created within specified time.
+ * @devnm: Device for which mdmon control socket should start.
+ *
+ * In foreground mode, when mdadm is trying to connect to control
+ * socket it is possible that the mdmon has not created it yet.
+ * Give some time to mdmon to create socket. Timeout set to 10 sec.
+ */
+int wait_for_mdmon_control_socket(char *devname)
```
mtkaczyk, line 1943: I see that the code in this function is a copy of the code from `connect_monitor`
In this form it cannot be accepted, we should not repeat same code. It must be optimized some way.

You can extract similar parts to helper functions, for example (not tested):
```c
mdadm_status_t get_mdmon_socket_file(char *devnm, char *buf)
{
        char *c;

	if (is_subarray(devname)) {
		devname++;
		c = strchr(devname, '/');
		if (!c)
			return MDADM_STATUS_ERROR;
               c++;
	} else
		c = devname;
	snprintf(buf, PATH_MAX, "%s/%s.sock", MDMON_DIR, c);
}
```

This is just a part of function, other places should be optimized too.

util.c
```diff
@@ -1932,6 +1932,55 @@ int mdmon_running(const char *devnm)
...
+	sprintf(&path[pos], ".sock");
+
+	sfd = socket(PF_LOCAL, SOCK_STREAM, 0);
+	if (sfd < 0)
+		return -1;
+
+	addr.sun_family = PF_LOCAL;
+	strcpy(addr.sun_path, path);
+
+	for (retry_count = 0; retry_count < 20; retry_count++) {
```
mtkaczyk, line 1970: Do we need 10 seconds? Previously, we determined that 2 seconds are enough.

mtkaczyk: 10 seconds will block it from being used in other places in the future, especially from udev context. Can we stay with 2 seconds as we previously determined?

util.c
shmsingh: I will try to optimize the similar code.

util.c
shmsingh: I had internal discussion with my team, and we considered a scenario where system is running heavy workload or is under heavy stress, in that case it is possible that the mdmon can take some time to create a socket. That's why I changed the time back to 10 seconds.

Also as far as I remember the udev kill a process/thread if it didn't get a response for more than 180 seconds. Please correct me if you are referring to something else.

util.c
shmsingh: @mtkaczyk I have pushed in the new changes. Please share your feedback.

FYI, while testing these changes, I noticed one thing that we might not need subarray check in case of container or specficially in case of wait_for_mdmon_control_socket() function context, because container name itself wont be in the subarray name format. Container name is in the format of "mdX" or "mdY" while subarray  can be of the format of "/md20/X" or "/md20/Y". 
I am still new to these terminologies or naming conventions, so I am not sure if the conclusion I have come to is correct or not.

util.c
mtkaczyk: Here a man:
https://man7.org/linux/man-pages/man7/udev.7.html
```
 IMPORT{type}
           Import a set of variables as device properties, depending on
           type:

           "program"
               Execute an external program specified as the assigned
               value and, if it returns successfully, import its output,
               which must be in environment key format. Path
               specification, command/argument separation, and quoting
               work like in RUN.

 {...}

           This can only be used for very short-running foreground
           tasks. For details see RUN.

```

It is not a case for create for sure, create is done from cmdline. But in a case that at some point I would like to reuse this logic in Incremental then I may get a problem, because we have:

```
ACTION!="remove", IMPORT{program}="BINDIR/mdadm --incremental --export $devnode --offroot $env{DEVLINKS}"
```
I can see that in `assemble_container_content` we are pinging mdmon in case of reshape so the question is if we should change that to the method you proposed. And here is why I would like to avoid long running task.

Udev is not strict what are "short-running foreground tasks". I expect seconds not minutes to be allowed. But this is still theoretical issue as we expect mdmon to be started fast.

So I would stay that if there is no real issue we are familiar with that requires to wait longer for mdmon, we should stay with 2 seconds unless you can really prove that more than 10 seconds is safe.

util.c
mtkaczyk: Thank you :)

util.c
mtkaczyk: > What if we pass timeout value as argument to the function? This way we can have different timeout value for different sceanrios.

Has some sense but we cannot expect from future developers to know what is reasonable. Also, please ensure that the research you are doing will not be forgotten. I want to respect the time you have spent on analyzing it.

So, please prepare some predefined values that developers may reuse without additional analysis. For example, by using `bool udev_frendly_wait` and defining long and shot time that is reasonable in your opinion. I will trust your judgement here.

 Sometimes `bool force` is used for similar purposes in mdadm. I have no strong preference for the name but please comment it well.

shmsingh: Lets do one thing, I will switch back to 2 second delay and I can add an error print if connect fails even after 2 seconds. 
And in future if we see connect failing in wait_for_mdmon_control_socket() for cases where we dont want it to fail then we can work on increasing the timeout limit along with all the fail safe for udev.

shmsingh: I have changed the sleep timeout to 2 second and have also added a debug print in case timout occur in wait_for_mdmon_control_socket().
Please review it.

mtkaczyk: Thanks, will look into it tomorrow!

shmsingh: A gentle reminder.

util.c
mtkaczyk: > FYI, while testing these changes, I noticed one thing that we might not need subarray check in case of container or specficially in case of wait_for_mdmon_control_socket() function context, because container name itself wont be in the subarray name format. Container name is in the format of "mdX" or "mdY" while subarray can be of the format of "/md20/X" or "/md20/Y". I am still new to these terminologies or naming conventions, so I am not sure if the conclusion I have come to is correct or not.

that is true, this is simplification for case when you are opening subarray only. This can be read from /sys/block/<subarray>/md/version. if you don't need this logic now fell free to remove. No need to keep dead code.

util.c
```diff
@@ -1932,6 +1932,58 @@ int mdmon_running(const char *devnm)
 	return 0;
 }
 
+/* wait_for_mdmon_control_socket() - Waits for mdmon control socket
```
mtkaczyk, line 1935: Please use kernel code style, generally it is:
```
/**
 * wait_for_mdmon_control_socket() - Waits for mdmon control socket.
 * @devnm: Device for which mdmon control socket should start.
 * 
 * In foreground mode, when mdadm is trying to connect to control
 * socket it is possible that the mdmon has not created it yet.
 *  Give some time to mdmon to create socket. Timeout set to 10 sec.
 */
```
Please read:
https://www.kernel.org/doc/html/v4.10/process/coding-style.html
for further reference.

util.c
```diff
@@ -1932,6 +1932,58 @@ int mdmon_running(const char *devnm)
...
+ * to be created within specified time.
+ * @devnm: Device for which mdmon control socket should start.
+ *
+ * In foreground mode, when mdadm is trying to connect to control
+ * socket it is possible that the mdmon has not created it yet.
+ * Give some time to mdmon to create socket. Timeout set to 10 sec.
+ */
+int wait_for_mdmon_control_socket(char *devname)
```
mtkaczyk, line 1943: after removing `subarray` determination, you can mark `devname` with const and possibly rename this to `const char *container_devnm`.

I'm trying to keep following approach for new code:
devname - means devnode path, like/dev/md127
devnm- means just device name md127

but either way it is fine for me, please try use const.

util.c
```diff
@@ -1932,6 +1932,58 @@ int mdmon_running(const char *devnm)
...
+		memcpy(name, devname, len);
+		name[len] = '\0';
+	} else {
+		len = strlen(devname);
+		memcpy(name, devname, len);
+		name[len] = '\0';
+	}
+	snprintf(path, sizeof(path), "%s/%s.sock", MDMON_DIR, name);
+	sfd = socket(PF_LOCAL, SOCK_STREAM, 0);
+	if (sfd < 0)
```
mtkaczyk, line 1966: we have `is_fd_valid` helper, you should it

util.c
```diff
@@ -1932,6 +1932,58 @@ int mdmon_running(const char *devnm)
...
+ * to be created within specified time.
+ * @devnm: Device for which mdmon control socket should start.
+ *
+ * In foreground mode, when mdadm is trying to connect to control
+ * socket it is possible that the mdmon has not created it yet.
+ * Give some time to mdmon to create socket. Timeout set to 10 sec.
+ */
+int wait_for_mdmon_control_socket(char *devname)
+{
+	char path[100], name[20];
```
mtkaczyk, line 1945: we have defines for sizes. Path should be `PATH_MAX` size. name, should be `MD_NAME_MAX`.

util.c
```diff
@@ -1932,6 +1932,58 @@ int mdmon_running(const char *devnm)
...
+			sleep_for(0, MSEC_TO_NSEC(200), true);
+			continue;
+		}
+		break;
+	}
+
+	if (rv < 0)
+		pr_err("Failed to connect to control socket.\n");
+	close(sfd);
+	return rv;
```
mtkaczyk, line 1984: Please consider using mdadm_status_t for errors instead of raw int as a return value. Outside this we are not interested in connect return status.  We just need information that it failed so I think that `MDADM_STATUS_ERROR` is fine

mtkaczyk (commented):
Hello, I looked into it again. I hope I answered all questions.
There are few new comments related to new style we are trying to keep.
Thanks!

shmsingh:
@mtkaczyk Thanks for your review comments. I have addressed all. Can you please review the updated patch. 
Thanks

util.c
```diff
@@ -1932,6 +1932,50 @@ int mdmon_running(const char *devnm)
...
+	struct sockaddr_un addr;
+	char path[PATH_MAX];
+
+	snprintf(path, PATH_MAX, "%s/%s.sock", MDMON_DIR, container_devnm);
+	sfd = socket(PF_LOCAL, SOCK_STREAM, 0);
+	if (!is_fd_valid(sfd))
+		return MDADM_STATUS_ERROR;
+
+	addr.sun_family = PF_LOCAL;
+	strcpy(addr.sun_path, path);
```
mtkaczyk, line 1960: Static analysis reported:
`string_overflow You might overrun the 108-character destination string addr.sun_path by writing 4096 characters from path`
please change it to `strncpy`, eventually `snprintf` (to ensure null ending`.

mtkaczyk (commented):
I testes possibly worst scenario, no hang observed:
```
[gklab-localhost ~/mtkaczyk_work/mdadm/mdadm-github]# mdadm -CR imsm -e imsm -n2 /dev/nvme[01]n1^C
[gklab-localhost ~/mtkaczyk_work/mdadm/mdadm-github]# ./mdadm -CR vol -l 0 -n2 /dev/nvme[01]n1 -z 10G
mdadm: /dev/nvme0n1 appears to be part of a raid array:
       level=container devices=0 ctime=Thu Jan  1 01:00:00 1970
mdadm: /dev/nvme1n1 appears to be part of a raid array:
       level=container devices=0 ctime=Thu Jan  1 01:00:00 1970
mdadm: Creating array inside imsm container /dev/md127
mdadm: array /dev/md/vol started.
[gklab-localhost ~/mtkaczyk_work/mdadm/mdadm-github]# ./mdadm -CR vol1 -l 0 -n2 /dev/nvme[01]n1 -z 10G
mdadm: attempting to create a second volume with size less then remaining space.
mdadm: /dev/nvme0n1 appears to be part of a raid array:
       level=container devices=0 ctime=Thu Jan  1 01:00:00 1970
mdadm: attempting to create a second volume with size less then remaining space.
mdadm: /dev/nvme1n1 appears to be part of a raid array:
       level=container devices=0 ctime=Thu Jan  1 01:00:00 1970
mdadm: Creating array inside imsm container /dev/md127
mdadm: array /dev/md/vol1 started.
[gklab-localhost ~/mtkaczyk_work/mdadm/mdadm-github]# cat /proc/mdstat
Personalities : [raid1] [raid0]
md125 : active raid0 nvme1n1[1] nvme0n1[0]
      20971520 blocks super external:/md127/1 128k chunks

md126 : active raid0 nvme1n1[1] nvme0n1[0]
      20971520 blocks super external:/md127/0 128k chunks

md127 : inactive nvme1n1[1](S) nvme0n1[0](S)
      2210 blocks super external:imsm

unused devices: <none>
```
LGTM but please fix static anlysis issue.

mtkaczyk:
Hey @shmsingh 
Thank you!
There is a lot of issues and other coding opportunities here, if you would like to try let me know. We are looking for developers to help us with mdadm maintenance.

We are working on creating custom github runner for mdadm and invest more into upstream testing, this is another another opportunity you might be interested. Feel free to reach me if you are interested.

For now merging.

mtkaczyk (approved):

shmsingh:
Hi @mtkaczyk , thanks for the merging. I am still new to mdadm. But I would love to help wherever I will be able to. Thanks

mtkaczyk:
You can try with #22 :)

PR #35, opened 2024-06-24
mdadm: Build.c fix coverity issues

ncroxon:
Event leaked_handle: Handle variable "bitmap_fd" going out of scope leaks the handle.

### DIFF
--- a/Build.c
+++ b/Build.c
@@ -168,13 +168,13 @@ int Build(struct mddev_ident *ident, struct mddev_dev *devlist, struct shape *s,
 				goto abort;
 			}
 		}
-		if (bitmap_fd >= 0) {
-			if (ioctl(mdfd, SET_BITMAP_FILE, bitmap_fd) < 0) {
-				pr_err("Cannot set bitmap file for %s: %s\n", chosen_name,
-				       strerror(errno));
-				goto abort;
-			}
+		if (ioctl(mdfd, SET_BITMAP_FILE, bitmap_fd) < 0) {
+			pr_err("Cannot set bitmap file for %s: %s\n", chosen_name,
+			       strerror(errno));
+			close(bitmap_fd);
+			goto abort;
 		}
+	close(bitmap_fd);
 	}
 	if (ioctl(mdfd, RUN_ARRAY, &param)) {
 		pr_err("RUN_ARRAY failed: %s\n", strerror(errno));


Build.c
```diff
@@ -40,7 +40,7 @@ int Build(struct mddev_ident *ident, struct mddev_dev *devlist, struct shape *s,
 	dev_t rdev;
 	int subdevs = 0, missing_disks = 0;
 	struct mddev_dev *dv;
-	int bitmap_fd;
+	int bitmap_fd = 0;
```
mtkaczyk, line 43: I don't think assignment is needed here.

ncroxon:
pushed a new version

ncroxon:
Mariusz, that assignment is required..

fatal: No names found, cannot describe anything.
Build.c: In function ‘Build’:
Build.c:192:9: error: ‘bitmap_fd’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
  192 |         close(bitmap_fd);
      |         ^~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors

Build.c
```diff
@@ -188,6 +189,7 @@ int Build(struct mddev_ident *ident, struct mddev_dev *devlist, struct shape *s,
 		pr_err("array %s built and started.\n", chosen_name);
 	wait_for(chosen_name, mdfd);
 	close(mdfd);
+	close(bitmap_fd);
```
bkucman, line 192: this close call, should be executed at the end of `if (s->bitmap_file) {` block, as it is not used outside of it.
also declaration should be moved to "if" block 
`int bitmap_fd = open(s->bitmap_file, O_RDWR);`

bkucman (commented):
hi,
it can be simplified for readability, bitmap, it is only needed inside in one "if" block

mtkaczyk:
> Mariusz, that assignment is required..
> 
> fatal: No names found, cannot describe anything. Build.c: In function ‘Build’: Build.c:192:9: error: ‘bitmap_fd’ may be used uninitialized in this function [-Werror=maybe-uninitialized] 192 | close(bitmap_fd); | ^~~~~~~~~~~~~~~~ cc1: all warnings being treated as errors

@ncroxon this is wrong then. according to this  it may close not opened fd. In that case:
- you should set it to -1 initially (0 is a valid fd number!)
- use `close_fd()` helper instead of `close()`. The helper routine does fd verification first.

I see that Błazej commented that we can keep that under  `if (s->bitmap_file) {` and that may simplify handling a little. Please consider moving it.

Thanks!

Build.c
```diff
@@ -188,10 +188,12 @@ int Build(struct mddev_ident *ident, struct mddev_dev *devlist, struct shape *s,
...
```
bkucman, line 191:
```suggestion
if (s->bitmap_file) {
		int bitmap_fd = open(s->bitmap_file, O_RDWR);
		if (bitmap_fd < 0) {
			int major = BITMAP_MAJOR_HI;
			bitmapsize = s->size >> 9; /* FIXME wrong for RAID10 */
			if (CreateBitmap(s->bitmap_file, 1, NULL,
					 s->bitmap_chunk, c->delay,
					 s->write_behind, bitmapsize, major)) {
				goto abort;
			}
			bitmap_fd = open(s->bitmap_file, O_RDWR);
			if (bitmap_fd < 0) {
				pr_err("%s cannot be opened.\n", s->bitmap_file);
				goto abort;
			}
		}
		if (ioctl(mdfd, SET_BITMAP_FILE, bitmap_fd) < 0) {
			pr_err("Cannot set bitmap file for %s: %s\n", chosen_name,
				 strerror(errno));
			close(bitmap_fd);
			goto abort;
		}
		close(bitmap_fd);
	}
```

bkucman: Hi, 
please look at suggested change above , this is what I mean in my previous comments
"bitmap_fd, it is only needed inside in one "if" block" and "this check "if (bitmap_fd >= 0) {" is redundant , previous "if" block in case of bitmap_fd < 0 will go to abort:"

bkucman (changes requested):

Build.c
ncroxon: added your code the patch.

Build.c
mtkaczyk: Please change it to -1 or do not assing value here.

mtkaczyk:
Hi @ncroxon,
The requested changes are not added, I reopened the conversations, please take a look.

Build.c
ncroxon: Strange... as it is in my version...  when I push a new version, it should be there.

ncroxon:
Pushed a new version..

mtkaczyk (approved):

bkucman (approved):

PR #36, opened 2024-06-25
CI: fix excluded files in checkpatch.conf

ktanska:
--exclude flag in checkpatch.conf is configured to work on directories or files but without using regexes. Remove regexes and add absolute file names, which should be excluded from checkpatch.


mtkaczyk (approved):

.checkpatch.conf
```diff
@@ -6,11 +6,19 @@
...
---exclude *.8
+--exclude CHANGELOG.md
+--exclude MAINTAINERS.md
+--exclude README.md
+--exclude md.4
+--exclude mdadm.8.in
+--exclude mdadm.conf.5.in
+--exclude mdmon.8
+--exclude raid6check.8
 --exclude test
```
bkucman, line 17: --exclude seems to only work for directories

ktanska:
Checkpatch.conf may exclude only directories for now.
I've submitted new issue to its maintainers, if they will extend checkpatch.conf flags and add possibility to exclude files I will send new patch to mdadm: [https://github.com/zephyrproject-rtos/zephyr/issues/74957](https://github.com/zephyrproject-rtos/zephyr/issues/74957)

bkucman (approved):

PR #39, opened 2024-06-25
mdadm: Create.c fix coverity issues

ncroxon:
* Event negative_returns: "fd" is passed to a parameter that cannot be negative. Which is set to -1 to start.

* Event open_fn: Returning handle opened by "open_dev_excl".
* Event var_assign: Assigning: "container_fd" = handle returned from "open_dev_excl(st->container_devnm)"
* Event leaked_handle: Handle variable "container_fd" going out of scope leaks the handle

### DIFF
--- a/Create.c
+++ b/Create.c
@@ -297,7 +297,7 @@ static int add_disk_to_super(int mdfd, struct shape *s, struct context *c,
 	if (st->ss->add_to_super(st, &info->disk, fd, dv->devname,
 				 dv->data_offset)) {
 		ioctl(mdfd, STOP_ARRAY, NULL);
-		close(fd);
+		close_fd(&fd);
 		return 1;
 	}
 	st->ss->getinfo_super(st, info, NULL);
@@ -1368,8 +1368,8 @@ int Create(struct supertype *st, struct mddev_ident *ident, int subdevs,
 	map_remove(&map, fd2devnm(mdfd));
 	map_unlock(&map);
 
-	if (mdfd >= 0)
-		close(mdfd);
+	close_fd(&mdfd);
+	close_fd(&container_fd);
 
 	dev_policy_free(custom_pols);
 	return 1;


Create.c
```diff
@@ -1371,6 +1371,9 @@ int Create(struct supertype *st, struct mddev_ident *ident, int subdevs,
 	if (mdfd >= 0)
 		close(mdfd);
 
+	if (container_fd >= 0)
```
mtkaczyk, line 1374: please just use `close_fd()`

Create.c
```diff
@@ -294,8 +294,8 @@ static int add_disk_to_super(int mdfd, struct shape *s, struct context *c,
 	}
 	if (fd >= 0)
 		remove_partitions(fd);
-	if (st->ss->add_to_super(st, &info->disk, fd, dv->devname,
-				 dv->data_offset)) {
+	if ((st->ss->add_to_super(st, &info->disk, fd, dv->devname,
+				 dv->data_offset)) && (!have_container)) {
```
mtkaczyk, line 298: From review perspective I cannot judge if it is safe. Did you tested it?

I understand that we set `fd = -1` if `have_container` but it skipps `ioctl(mdfd, STOP_ARRAY, NULL);` is a logical change. Perhaps you can just use `close_fd(&fd)` and avoid this?

Create.c
ncroxon: fixed

Create.c
```diff
@@ -296,8 +296,7 @@ static int add_disk_to_super(int mdfd, struct shape *s, struct context *c,
 		remove_partitions(fd);
 	if (st->ss->add_to_super(st, &info->disk, fd, dv->devname,
 				 dv->data_offset)) {
-		ioctl(mdfd, STOP_ARRAY, NULL);
```
mtkaczyk, line 299: ioctl call must stay unless you can prove that it is necessary.

Create.c
```diff
@@ -1371,6 +1370,8 @@ int Create(struct supertype *st, struct mddev_ident *ident, int subdevs,
 	if (mdfd >= 0)
 		close(mdfd);
 
+	close_fd(&container_fd);
```
mtkaczyk, line 1373: by chance, you can change this mdfd check upper :)

mtkaczyk (changes requested):

Create.c
ncroxon: I put it back... I haven't proven that it is unnecessary.

Create.c
ncroxon: @mtkaczyk what do you mean here?

mtkaczyk:
```c
	close_fd(&mdfd);
	close_fd(&container_fd);
```

ncroxon: gotcha.

mtkaczyk (approved):

PR #42, opened 2024-07-02
CI: use prepared checkpatch.conf file only for GH actions

ktanska:
Configuration file .checkpatch.conf is working properly only with GH actions, because flags from GH plugin are used there. This file shall not be placed in main repo directory, because it causes errors while using checkpatch from Linux. Add step to review.yml to copy this file before checkpatch action is started.


mku514k (approved):
LGTM

mtkaczyk (approved):

PR #45, opened 2024-07-04
Compiler flags

ktanska:
It is essential to avoid vulnerabilities in code as much
as possible using safe compilation flags. It is easier if
they are added to the Makefile and applied during compilation.
Add new gcc flags and make them configurable, because they
may not be supported for some compilers.
Unset preconfigured FORTIFY_SOURCE, which caused build errors
when built-in flag was used with the same value.
Fix error reported when those additional flags were used.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,7 @@
 
 # define "CXFLAGS" to give extra flags to CC.
 # e.g.  make CXFLAGS=-O to optimise
-CXFLAGS ?=-O2 -D_FORTIFY_SOURCE=2
+CXFLAGS ?=-O2 -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE
 TCC = tcc
 UCLIBC_GCC = $(shell for nm in i386-uclibc-linux-gcc i386-uclibc-gcc; do which $$nm > /dev/null && { echo $$nm ; exit; } ; done; echo false No uclibc found )
 #DIET_GCC = diet gcc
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -76,6 +76,27 @@ ifeq ($(origin STRINGOPOVERFLOW), undefined)
...
+endif
+
+ifeq ($(origin NODELETENULLPOINTER), undefined)
+	NODELETENULLPOINTER := $(shell $(CC) -Q --help=optimizers 2>&1 | grep "delete-null-pointer-checks" | wc -l)
+	ifneq "$(NODELETENULLPOINTER)"  "0"
+	CWFLAGS += -fno-delete-null-pointer-checks
+	endif
+endif
+
+ifeq ($(origin WRAPV), undefined)
```
mtkaczyk, line 93: I don't like how our makefile is expanding. It is right time to start considering automake/cmake.

mdadm.h
```diff
@@ -2021,6 +2021,10 @@ enum r0layout {
 #define PATH_MAX	4096
 #endif
 
+/* Define length of char array which will contain int converted.
+ */
+#define INT_MAX_LEN 12
```
mtkaczyk, line 2026: The name is not right, it should be something like `INT_DEC_STR_MAX`

This is system dependent setting because `sizeof(int)` may not be 32. I would prefer:
```c
/* The max string length necessary for decimal conversion, cannot be longer than count of bits */
#define INT_2_DEC_STR_MAX sizeof(int) * 8
// Here comes other examples, just FYI
/* For binary conversion, in the worst we need one char more */
#define INT_2_BIN_STR_MAX INT_2_DEC_STR_MAX + 1
/* One hex sign represent one byte */
#define INT_2_HEX_STR_MAX sizeof(int) + 1
```
Using hex is more convenient but we have decimal based API..

mtkaczyk (changes requested):

Makefile
```diff
@@ -30,7 +30,7 @@
 
 # define "CXFLAGS" to give extra flags to CC.
 # e.g.  make CXFLAGS=-O to optimise
-CXFLAGS ?=-O2 -D_FORTIFY_SOURCE=2
+CXFLAGS ?=-O2 -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=3
```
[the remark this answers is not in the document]
peluse: Excellent observation @mtkaczyk! My gut tells me the impact for mdadm would not be noticeable though, maybe some quick testing would make sense?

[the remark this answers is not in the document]
mtkaczyk: Well, it is tested on production starting with Redhat 9.2 (or even earlier)
No issues detected, so I assume it is safe for mdadm.

Makefile
ktanska: I've started discussion #62

mtkaczyk (approved):

PR #46, opened 2024-07-05
mdstat: Rework mdstat external ararys handling

mtkaczyk:
To avoid repeating `mdstat_read()` in `IncrementalRemove()`, new function `mdstat_find_by_member_name()` has been proposed. With that, `IncrementalRemove()` handles own copy of mdstat content and there is no need to repeat reading for external stop.

Additionally, It proposed few helper to avoid repeating `ent->metadata_version` checks across code.

Tested by:
- `mdadm -If nvme0n1` - This involves testing `mdstat_find_by_member_name()`
- `mdadm -G -l0 /dev/md126` - This involves `mdadm_by_subdev()`

### DIFF
--- a/Assemble.c
+++ b/Assemble.c
@@ -114,14 +114,11 @@ static int is_member_busy(char *metadata_version)
 	int busy = 0;
 
 	for (ent = mdstat; ent; ent = ent->next) {
-		if (ent->metadata_version == NULL)
+		if (!is_mdstat_ent_subarray(ent))
 			continue;
-		if (strncmp(ent->metadata_version, "external:", 9) != 0)
-			continue;
-		if (!is_subarray(&ent->metadata_version[9]))
-			continue;
+
 		/* Skip first char - it can be '/' or '-' */
-		if (strcmp(&ent->metadata_version[10], metadata_version+1) == 0) {
+		if (strcmp(&ent->metadata_version[10], metadata_version + 1) == 0) {
 			busy = 1;
 			break;
 		}
--- a/Incremental.c
+++ b/Incremental.c
@@ -1686,12 +1686,13 @@ static void remove_from_member_array(struct mdstat_ent *memb,
  */
 int IncrementalRemove(char *devname, char *id_path, int verbose)
 {
-	int mdfd;
-	int rv = 0;
-	struct mdstat_ent *ent;
+	struct mdstat_ent *ent = NULL;
+	char buf[SYSFS_MAX_BUF_SIZE];
+	struct mdstat_ent *mdstat;
 	struct mddev_dev devlist;
 	struct mdinfo mdi;
-	char buf[SYSFS_MAX_BUF_SIZE];
+	int rv = 1;
+	int mdfd;
 
 	if (!id_path)
 		dprintf("incremental removal without --path <id_path> lacks the possibility to re-add new device in this port\n");
@@ -1700,16 +1701,25 @@ int IncrementalRemove(char *devname, char *id_path, int verbose)
 		pr_err("incremental removal requires a kernel device name, not a file: %s\n", devname);
 		return 1;
 	}
-	ent = mdstat_by_component(devname);
+
+	mdstat = mdstat_read(0, 0);
+	if (!mdstat) {
+		pr_err("Cannot read /proc/mdstat file, aborting\n");
+		return 1;
+	}
+
+	ent = mdstat_find_by_member_name(mdstat, devname);
 	if (!ent) {
 		if (verbose >= 0)
 			pr_err("%s does not appear to be a component of any array\n", devname);
-		return 1;
+		goto out;
 	}
+
 	if (sysfs_init(&mdi, -1, ent->devnm)) {
 		pr_err("unable to initialize sysfs for: %s\n", devname);
-		return 1;
+		goto out;
 	}
+
 	mdfd = open_dev_excl(ent->devnm);
 	if (is_fd_valid(mdfd)) {
 		close_fd(&mdfd);
@@ -1725,8 +1735,7 @@ int IncrementalRemove(char *devname, char *id_path, int verbose)
 	if (mdfd < 0) {
 		if (verbose >= 0)
 			pr_err("Cannot open array %s!!\n", ent->devnm);
-		free_mdstat(ent);
-		return 1;
+		goto out;
 	}
 
 	if (id_path) {
@@ -1741,16 +1750,13 @@ int IncrementalRemove(char *devname, char *id_path, int verbose)
 	devlist.devname = devname;
 	devlist.disposition = 'I';
 	/* for a container, we must fail each member array */
-	if (ent->metadata_version &&
-	    strncmp(ent->metadata_version, "external:", 9) == 0) {
-		struct mdstat_ent *mdstat = mdstat_read(0, 0);
+	if (is_mdstat_ent_external(ent)) {
 		struct mdstat_ent *memb;
 		for (memb = mdstat ; memb ; memb = memb->next) {
 			if (is_container_member(memb, ent->devnm))
 				remove_from_member_array(memb,
 					&devlist, verbose);
 		}
-		free_mdstat(mdstat);
 	} else {
 		/*
 		 * This 'I' incremental remove is a try-best effort,
@@ -1765,7 +1771,8 @@ int IncrementalRemove(char *devname, char *id_path, int verbose)
 	rv = Manage_subdevs(ent->devnm, mdfd, &devlist,
 			    verbose, 0, UOPT_UNDEFINED, 0);
 
-	close(mdfd);
-	free_mdstat(ent);
+	close_fd(&mdfd);
+out:
+	free_mdstat(mdstat);
 	return rv;
 }
--- a/Manage.c
+++ b/Manage.c
@@ -278,10 +278,8 @@ int Manage_stop(char *devname, int fd, int verbose, int will_retry)
 		 */
 		mds = mdstat_read(0, 0);
 		for (m = mds; m; m = m->next)
-			if (m->metadata_version &&
-			    strncmp(m->metadata_version, "external:", 9)==0 &&
-			    metadata_container_matches(m->metadata_version+9,
-						       devnm)) {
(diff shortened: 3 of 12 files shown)


Incremental.c
```diff
@@ -1700,16 +1701,24 @@ int IncrementalRemove(char *devname, char *id_path, int verbose)
 		pr_err("incremental removal requires a kernel device name, not a file: %s\n", devname);
 		return 1;
 	}
-	ent = mdstat_by_component(devname);
+	mdstat = mdstat_read(0, 0);
+	if (!mdstat) {
+		pr_err("Cannot read /proc/mdstat file, aborting");
```
mku514k, line 1706: Looks like new line missing.

Incremental.c
```diff
@@ -1766,6 +1772,6 @@ int IncrementalRemove(char *devname, char *id_path, int verbose)
 			    verbose, 0, UOPT_UNDEFINED, 0);
 
 	close(mdfd);
-	free_mdstat(ent);
+	free_mdstat(mdstat);
```
mku514k, line 1775: Shouldn't mdstat be freed with  each return?

mdstat.c
```diff
@@ -110,6 +110,28 @@ static int add_member_devname(struct dev_member **m, char *name)
...
+
+	if (ent == el) {
+		*list_head = ent->next;
+		ent->next = NULL;
+		return;
+	}
+
+	while (ent) {
+		if (ent->next == el) {
+			ent->next = el->next;
```
mku514k, line 126: Shouldn't this be assigned to `*list_head`?

mdstat.c
```diff
@@ -110,6 +110,28 @@ static int add_member_devname(struct dev_member **m, char *name)
...
+{
+	struct mdstat_ent *ent = *list_head;
+
+	if (ent == el) {
+		*list_head = ent->next;
+		ent->next = NULL;
+		return;
+	}
+
+	while (ent) {
```
mku514k, line 124: This is just a thought, but maybe we could use "do while" here to get rid of this first check.

Manage.c
```diff
@@ -278,10 +278,8 @@ int Manage_stop(char *devname, int fd, int verbose, int will_retry)
 		 */
 		mds = mdstat_read(0, 0);
 		for (m = mds; m; m = m->next)
-			if (m->metadata_version &&
-			    strncmp(m->metadata_version, "external:", 9)==0 &&
-			    metadata_container_matches(m->metadata_version+9,
-						       devnm)) {
+			if (is_mdstat_ent_external(m) &&
+			    metadata_container_matches(m->metadata_version + 9, devnm)) {
```
mku514k, line 282: This patch might be an opportunity to get rid of "+9" magic number and use a define (like VERSION_SHIFT or smth.), but I leave it up to you.

mdstat.c
mtkaczyk: Reworked

mdstat.c
mtkaczyk: No, first element has special handling.

Incremental.c
mtkaczyk: Yes it should

Manage.c
mtkaczyk: I tried.. it is too big task. I  requires to rework a lot of code and it is not simple sometimes, I need to be extra careful (and I may not be able to test everything.

Let me finish this for now!

mdmon.c
```diff
@@ -394,9 +394,7 @@ int main(int argc, char *argv[])
 		/* launch an mdmon instance for each container found */
 		mdstat = mdstat_read(0, 0);
 		for (e = mdstat; e; e = e->next) {
-			if (e->metadata_version &&
-			    strncmp(e->metadata_version, "external:", 9) == 0 &&
-			    !is_subarray(&e->metadata_version[9])) {
+			if (is_mdstat_ent_subarray(e)) {
```
bkucman, line 397: incorrect check,
is_mdstat_ent_subarray expect volume to be subarray, but this if statement expect oposite state

config.c
```diff
@@ -353,35 +353,41 @@ struct mddev_dev *load_partitions(void)
...
 	struct mdstat_ent *mdstat = mdstat_read(0, 0);
+	struct mddev_dev *dev_list = NULL;
+	struct map_ent *map_list = NULL;
 	struct mdstat_ent *ent;
-	struct mddev_dev *d;
-	struct mddev_dev *rv = NULL;
-	struct map_ent *map = NULL, *me;
 
 	if (!mdstat)
 		return NULL;
```
bkucman, line 361: this check seems redundant, in case of mdstat == NULL, NULL will be returned through dev_list
`if (!mdstat)
	return NULL;
`

Incremental.c
```diff
@@ -1725,8 +1735,7 @@ int IncrementalRemove(char *devname, char *id_path, int verbose)
 	if (mdfd < 0) {
```
bkucman, line 1735: is_fd_valid

Assemble.c
```diff
@@ -114,12 +114,9 @@ static int is_member_busy(char *metadata_version)
...
-		if (ent->metadata_version == NULL)
-			continue;
-		if (strncmp(ent->metadata_version, "external:", 9) != 0)
-			continue;
-		if (!is_subarray(&ent->metadata_version[9]))
+		if (!is_mdstat_ent_subarray(ent))
 			continue;
+
 		/* Skip first char - it can be '/' or '-' */
 		if (strcmp(&ent->metadata_version[10], metadata_version+1) == 0) {
```
bkucman, line 121:
```suggestion
		if (strcmp(&ent->metadata_version[10], metadata_version + 1) == 0) {
```

bkucman (changes requested):
One typo in commit title
`ararys ->arrays'

Manage.c
mtkaczyk: Simple replacement of `+ 9` by `+ EXTERNAL_INDICATOR_LEN` makes it only worse (I personally don't like it). The right way to do it is to propose separate family to handle `metadata version`. I tried  once and I ended with 5 inlines and it was not enough, it will lead me to rework almost each mdstat usage. This must be done separately.
This time, I want to end mdstat lists rework and I don't want to complicate this task more. It is 3th story started by me on on ``--set-faulty`` and ``--remove`` problem.

Also, to propose "correct" naming it all must be done together. I had a problem how to name `/md127/1` substring extracted from `metadata_version`. 

Thanks for this suggestion but I can't for now :(

mdstat.c
```diff
@@ -382,61 +430,70 @@ int mddev_busy(char *devnm)
 	return me != NULL;
 }
 
-struct mdstat_ent *mdstat_by_component(char *name)
+/**
+ * mdstat_find_by_member_devnm()- Return first array or external container with member device.
+ * @mdstat: Preloaded mdstat to iterate over.
```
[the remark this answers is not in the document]
mtkaczyk: If you are putting NULL as mdstat then NULL is returned, that is fair.
Should I mention it or assert? I don't think it matters. Caller must be prepared that NULL may be returned.

mdstat.c
```diff
@@ -382,61 +430,70 @@ int mddev_busy(char *devnm)
...
+	free_mdstat(mdstat);
+
+	return ent;
+}
+
 struct mdstat_ent *mdstat_by_subdev(char *subdev, char *container)
 {
 	struct mdstat_ent *mdstat = mdstat_read(0, 0);
-	struct mdstat_ent *ent = NULL;
+	struct mdstat_ent *ent;
```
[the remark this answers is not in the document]
mtkaczyk: yes, theoretically mdstat may be NULL so I can keep it assigned.

mdmon.c
mtkaczyk: good catch! As I said, it requires to be extra careful and I failed :)

config.c
mtkaczyk: Right, fixed.

Incremental.c
mtkaczyk: will be reworked in IncrementalRemove PR, I started with, not here.

Assemble.c
mtkaczyk: ok, applied.

Manage.c
bkucman: Ack :)

mdstat.c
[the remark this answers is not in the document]
bkucman: Got it.

mdmon.c
```diff
@@ -394,9 +394,7 @@ int main(int argc, char *argv[])
 		/* launch an mdmon instance for each container found */
 		mdstat = mdstat_read(0, 0);
 		for (e = mdstat; e; e = e->next) {
-			if (e->metadata_version &&
-			    strncmp(e->metadata_version, "external:", 9) == 0 &&
-			    !is_subarray(&e->metadata_version[9])) {
+			if (is_mdstat_ent_external(e) && !is_mdstat_ent_subarray(e)) {
```
bkucman, line 397: `!is_mdstat_ent_subarray(e)` introduce redundancy by calling is_mdstat_ent_external(e) second time
so maybe just use `'is_subarray(ent->metadata_version + 9)`
unless you want to avoid using the magic number again ;)

bkucman (changes requested):

bkucman (approved):

Incremental.c
mku514k: Are you sure "\n" is not needed here? You marked it as resolved without comment.

Incremental.c
```diff
@@ -1700,16 +1701,25 @@ int IncrementalRemove(char *devname, char *id_path, int verbose)
 		pr_err("incremental removal requires a kernel device name, not a file: %s\n", devname);
 		return 1;
 	}
-	ent = mdstat_by_component(devname);
+
+	mdstat = mdstat_read(0, 0);
+	if (!mdstat) {
+		pr_err("Cannot read /proc/mdstat file, aborting");
```
mku514k, line 1707: Are you sure "\n" is not needed here? You marked it as resolved without comment.

Incremental.c
mtkaczyk: Right, I missed it thanks!

PR #49, opened 2024-07-10
mdadm Monitor.c fix coverity issues

ncroxon:

* Event check_return: Calling "fcntl(fd, 2, 1)" without checking return value. This library function may fail and return an error code.

* Dereferencing "sl", which is known to be "NULL".

* Event fixed_size_dest: You might overrun the 32-character fixed-size string "devnm" by copying "tmp" without checking the length.


Monitor.c
```diff
@@ -782,7 +782,9 @@ static int check_array(struct state *st, struct mdstat_ent *mdstat,
 	if (!is_container && !md_array_active(fd))
 		goto disappeared;
 
-	fcntl(fd, F_SETFD, FD_CLOEXEC);
+	if (fcntl(fd, F_SETFD, FD_CLOEXEC) < 0)
+		goto disappeared;
```
bkucman, line 786: as fcntl can fail for many reasons, I would rather close the funtcion by "out"

bkucman (changes requested):

bkucman (changes requested):
Just one more small correction, I didn't notice it before.
Please add a colon after "mdadm" in the commit title.

ncroxon:
> Just one more small correction, I didn't notice it before. Please add a colon after "mdadm" in the commit title.

done and pushed.

bkucman (approved):

mtkaczyk (approved):

PR #52, opened 2024-07-10
mdadm managemon.c fix coverity issues

ncroxon:

* Event check_return: Calling "fcntl(fd, 4, fl)" without checking return value. This library function may fail and return an error code.

* Event check_after_deref: Null-checking "new" suggests that it may be null, but it has already been dereferenced on all paths leading to the check.

### DIFF
--- a/managemon.c
+++ b/managemon.c
@@ -776,10 +776,8 @@ static void manage_new(struct mdstat_ent *mdstat,
 
 error:
 	pr_err("failed to monitor %s\n", mdstat->metadata_version);
-	if (new) {
-		new->container = NULL;
-		free_aa(new);
-	}
+	new->container = NULL;
+	free_aa(new);
 	if (mdi)
 		sysfs_free(mdi);
 }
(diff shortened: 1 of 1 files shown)


bkucman (changes requested):
Hi Nigel,
Please add a colon after "mdadm" in the commit title.
Overall LGTM.

Thanks.

managemon.c
```diff
@@ -871,7 +869,10 @@ void read_sock(struct supertype *container)
 
 	fl = fcntl(fd, F_GETFL, 0);
 	fl |= O_NONBLOCK;
-	fcntl(fd, F_SETFL, fl);
+	if (fcntl(fd, F_SETFL, fl) < 0) {
```
mtkaczyk, line 872: `fnctl` is called two times but we are checking the result only on last attempt. I checked the man and it it not quarented that `F_GETFL` is always successful. We should check this status too.

mtkaczyk (changes requested):
LGTM but please check status of first `fnctl` in read_sock.

mtkaczyk:
Hey @ncroxon,
Checkpatch job if failing. Please fix it here and in other pulls.
```
ERROR:ASSIGN_IN_IF: do not use assignment in if condition
#42: FILE: managemon.c:870:
+ if ((fl = fcntl(fd, F_GETFL, 0)) < 0) {
Error: ERROR:ASSIGN_IN_IF: do not use assignment in if condition
```

Thanks!

ncroxon:
Fixed and pushed.

no issues with checkpatch

$ /scripts/checkpatch.pl 0001-mdadm-managemon.c-fix-coverity-issues.patch
total: 0 errors, 0 warnings, 28 lines checked

0001-mdadm-managemon.c-fix-coverity-issues.patch has no obvious style 
problems and is ready for submission.

bkucman (approved):

ktanska (approved):

PR #54, opened 2024-07-15
mdadm: block leading dot

mtkaczyk:
Do not allow to use '.' on first place for mddevice name. It improves user experience because there is suggestion that MD device might be hidden.

```
# ./mdadm -CR .imsm1 -e imsm -n2 /dev/nvme[34]n1
mdadm: Value ".imsm1" cannot be set as devname. Reason: Leading dot forbidden.

```

### DIFF
--- a/config.c
+++ b/config.c
@@ -188,8 +188,36 @@ inline void ident_init(struct mddev_ident *ident)
 	ident->uuid_set = 0;
 }
 
+/** ident_check_name() - helper function to verify name.
+ * @name: name to check.
+ * @prop_name: the name of the property it is validated against, used for logging.
+ * @cmdline: context dependent actions.
+ *
+ * @name must follow name's criteria, be POSIX compatible and does not have leading dot.
+ */
+static mdadm_status_t ident_check_name(const char *name, const char *prop_name, const bool cmdline)
+{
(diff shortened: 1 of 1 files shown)


config.c
```diff
@@ -242,12 +242,17 @@ mdadm_status_t _ident_set_devname(struct mddev_ident *ident, const char *devname
 	else
 		name = devname;
 
-	if (is_name_posix_compatible(name) == false) {
+	if (has_leading_dot(name)) {
```
bkucman, line 245: I think it would be worth using the same check order in _ident_set_devname and _ident_set_name
Currently it looks like:
_ident_set_devname:
- has_leading_dot
- is_name_posix_compatible
- is_string_lq

_ident_set_name:
- is_string_lq
- has_leading_dot
- is_name_posix_compatible

bkucman (changes requested):
Hi,
Please add to commit title that is about name/mddevice name

Thanks.

config.c
mtkaczyk: Right. It is good time to add helper. See updated.

bkucman:
> Hi, Please add to commit title that is about name/mddevice name
> 
> Thanks.

still valid

bkucman (changes requested):

bkucman (approved):

PR #55, opened 2024-07-16
mdadm: msg.c fix coverity issues

ncroxon:

* Event check_return: Calling "fcntl(sfd, 4, fl)" without checking return value. This library function may fail and return an error code.

### DIFF
--- a/msg.c
+++ b/msg.c
@@ -176,8 +176,15 @@ int connect_monitor(char *devname)
 	}
 
 	fl = fcntl(sfd, F_GETFL, 0);
+	if (fl < 0) {
+		close(sfd);
+		return -1;
+	}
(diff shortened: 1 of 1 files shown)


msg.c
```diff
@@ -177,7 +177,10 @@ int connect_monitor(char *devname)
 
 	fl = fcntl(sfd, F_GETFL, 0);
```
mtkaczyk, line 178: Please check status of first `fnctl` call

mtkaczyk (changes requested):
LGTM but Please check first `fnctl` call.

bkucman (approved):

mtkaczyk (approved):

PR #57, opened 2024-07-18
mdadm: util.c fix coverity issues

ncroxon:

* Event check_return: Calling "open" without checking return value
* Event check_return: Calling "lseek(fd, sector_size, 0)" without checking return value.
* Event leaked_handle: Handle variable "fd" going out of scope leaks the handle.
* Event leaked_storage: Variable "dir" going out of scope leaks the storage it points to.
* Event fixed_size_dest: You might overrun the 32-character fixed-size string "st->devnm" by copying "_devnm" without checking the length.
* Event fixed_size_dest: You might overrun the 32-character fixed-size string "container" by copying "dev" without checking the length.

### DIFF
--- a/util.c
+++ b/util.c
@@ -1253,7 +1253,7 @@ struct supertype *super_by_fd(int fd, char **subarrayp)
 			*subarray++ = '\0';
 			subarray = xstrdup(subarray);
 		}
-		strcpy(container, dev);
+		snprintf(container, sizeof(container), "%s", dev);
 		sysfs_free(sra);
 		sra = sysfs_read(-1, container, GET_VERSION);
 		if (sra && sra->text_version[0])
@@ -1430,7 +1430,8 @@ static int get_gpt_last_partition_end(int fd, unsigned long long *endofpart)
 	/* skip protective MBR */
 	if (!get_dev_sector_size(fd, NULL, &sector_size))
 		return 0;
-	lseek(fd, sector_size, SEEK_SET);
+	if (lseek(fd, sector_size, SEEK_SET) == -1L)
+		return 0;
 	/* read GPT header */
 	if (read(fd, &gpt, 512) != 512)
 		return 0;
@@ -1451,7 +1452,8 @@ static int get_gpt_last_partition_end(int fd, unsigned long long *endofpart)
 	part = (struct GPT_part_entry *)buf;
 
 	/* set offset to third block (GPT entries) */
-	lseek(fd, sector_size*2, SEEK_SET);
+	if (lseek(fd, sector_size*2, SEEK_SET) == -1L)
+		return 0;
 	for (part_nr = 0; part_nr < all_partitions; part_nr++) {
 		/* read partition entry */
 		if (read(fd, buf, entry_size) != (ssize_t)entry_size)
@@ -1486,7 +1488,8 @@ static int get_last_partition_end(int fd, unsigned long long *endofpart)
 
 	BUILD_BUG_ON(sizeof(boot_sect) != 512);
 	/* read MBR */
-	lseek(fd, 0, 0);
+	if (lseek(fd, 0, 0) == -1L)
+		goto abort;
 	if (read(fd, &boot_sect, 512) != 512)
 		goto abort;
 
@@ -1725,7 +1728,7 @@ int open_subarray(char *dev, char *subarray, struct supertype *st, int quiet)
 			       dev);
 		goto close_fd;
 	}
-	strcpy(st->devnm, _devnm);
+	snprintf(st->devnm, sizeof(st->devnm), "%s", _devnm);
 
 	mdi = sysfs_read(fd, st->devnm, GET_VERSION|GET_LEVEL);
 	if (!mdi) {
@@ -2303,14 +2306,16 @@ void manage_fork_fds(int close_all)
 {
 	DIR *dir;
 	struct dirent *dirent;
+	int fd = open("/dev/null", O_RDWR);
 
-	close(0);
-	open("/dev/null", O_RDWR);
-
+	if (is_fd_valid(fd)) {
+		dup2(fd, 0);
 #ifndef DEBUG
 	dup2(0, 1);
 	dup2(0, 2);
+	close_fd(&fd);
 #endif
+	}
 
 	if (close_all == 0)
 		return;
@@ -2329,8 +2334,10 @@ void manage_fork_fds(int close_all)
 
 		fd = strtol(dirent->d_name, NULL, 10);
 		if (fd > 2)
-			close(fd);
+			close_fd(&fd);
 	}
+	closedir(dir);
+	return;
 }
 
 /* In a systemd/udev world, it is best to get systemd to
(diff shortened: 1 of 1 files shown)


util.c
```diff
@@ -1253,7 +1253,7 @@ struct supertype *super_by_fd(int fd, char **subarrayp)
 			*subarray++ = '\0';
 			subarray = xstrdup(subarray);
 		}
-		strcpy(container, dev);
+		strncpy(container, dev, sizeof(container) - 1);
```
mtkaczyk, line 1256: please use `snprintf` for consistency.

util.c
```diff
@@ -1725,7 +1728,7 @@ int open_subarray(char *dev, char *subarray, struct supertype *st, int quiet)
 			       dev);
 		goto close_fd;
 	}
-	strcpy(st->devnm, _devnm);
+	strncpy(st->devnm, _devnm, sizeof(st->devnm) - 1);
```
mtkaczyk, line 1731: same here, please use snprintf.

util.c
```diff
@@ -2303,21 +2306,27 @@ void manage_fork_fds(int close_all)
 {
 	DIR *dir;
 	struct dirent *dirent;
+	int fd;
 
 	close(0);
-	open("/dev/null", O_RDWR);
+	fd = open("/dev/null", O_RDWR);
+	if (fd < 0)
```
mtkaczyk, line 2313: This is a way to redirect streams to `/dev/null`, if `/dev/null` is not achievable, we have to close them (I mean 0, 1, 2).
If should not be a real case but we need to handle it.

util.c
```diff
@@ -2303,21 +2306,27 @@ void manage_fork_fds(int close_all)
...
+		return;
 
 #ifndef DEBUG
 	dup2(0, 1);
 	dup2(0, 2);
 #endif
 
-	if (close_all == 0)
+	if (close_all == 0) {
+		close_fd(&fd);
```
mtkaczyk, line 2322: No, it must stay opened. It is redirection to /dev/null

util.c
```diff
@@ -2379,11 +2390,13 @@ void reopen_mddev(int mdfd)
 	 */
 	char *devnm;
 	int fd;
+
 	devnm = fd2devnm(mdfd);
 	close(mdfd);
 	fd = open_dev(devnm);
 	if (fd >= 0 && fd != mdfd)
 		dup2(fd, mdfd);
+	close_fd(&fd);
```
mtkaczyk, line 2399: If we assigned same descriptor number as `mdfd` had, we are closing it now - it leads to issues.
it should be only closed in if number is different.

Well, this function is bad, there is no error handling, but looks like it is working.

mtkaczyk (changes requested):
HI Nigel,
Please take a look into my comments.

util.c
```diff
@@ -2379,11 +2392,14 @@ void reopen_mddev(int mdfd)
 	 */
 	char *devnm;
 	int fd;
+
 	devnm = fd2devnm(mdfd);
 	close(mdfd);
 	fd = open_dev(devnm);
-	if (fd >= 0 && fd != mdfd)
+	if (fd >= 0 && fd != mdfd) {
```
bkucman, line 2399: is_fd_valid() && fd != mdfd

bkucman (changes requested):
Hi,  
please fix issues indicated by checkpatch and and a few comments from me.

ncroxon:
fixed the issues.

util.c
```diff
@@ -1253,7 +1253,7 @@ struct supertype *super_by_fd(int fd, char **subarrayp)
 			*subarray++ = '\0';
 			subarray = xstrdup(subarray);
 		}
-		strcpy(container, dev);
+		snprintf(container, sizeof(container) - 1, "%s", dev);
```
mtkaczyk, line 1256: with `snprintf`,  sizeof(container) - 1 is not needed. You can use  sizeof(container)

util.c
```diff
@@ -1725,7 +1728,7 @@ int open_subarray(char *dev, char *subarray, struct supertype *st, int quiet)
 			       dev);
 		goto close_fd;
 	}
-	strcpy(st->devnm, _devnm);
+	snprintf(st->devnm, sizeof(st->devnm) - 1, "%s", _devnm);
```
mtkaczyk, line 1731: same here, please use sizeof(st->devnm)

util.c
```diff
@@ -2379,11 +2391,13 @@ void reopen_mddev(int mdfd)
 	 */
 	char *devnm;
 	int fd;
+
```
mtkaczyk, line 2394: I think it is right way to make it. We have to keep mdfd in case if we cannot reopen the device, so it is better to close in only if we successed to reopen mddev.

```c
/**
 * reopen_mddev() - Reopen without O_EXCL, but keep the same fd numer.
 *@mdfd:  file descriptor with O_EXCL
 *
 * If it cannot reopen the device then mdfd is not reopened.
 */

void reopen_mddev(int mdfd)
{
	char *devnm = fd2devnm(mdfd);
	int fd = open_dev(devnm);

	if (!is_fd_valid(fd))
		return;

	dup2(fd, mdfd);

	close_fd(&fd);
}
```

util.c
```diff
@@ -2303,22 +2306,27 @@ void manage_fork_fds(int close_all)
...
 	struct dirent *dirent;
+	int fd;
 
 	close(0);
-	open("/dev/null", O_RDWR);
+	fd = open("/dev/null", O_RDWR);
+	if (!is_fd_valid(fd))
+		pr_err("open /dev/null failed.\n");
+	if (fd > 2)
+		close_fd(&fd);
```
mtkaczyk, line 2317: If we cannot open /dev/null (theoretical case) we should not close streams. What about?

```c
/* Close all opened descriptors if needed and redirect streams to /dev/null. If we cannot open
 * /dev/null then do not change them.
 * For debug purposed, leave STDOUT and STDERR untouched
 * Returns:
 *	1- if any error occurred
 *	0- otherwise
 */
void manage_fork_fds(int close_all)
{
	DIR *dir;
	struct dirent *dirent;
	int fd = open("/dev/null", O_RDWR);

	if (is_fd_valid(fd)) {
		dup(fd, 0)
#ifndef DEBUG
		dup2(fd, 1);
		dup2(fd, 2);
		close_fd(&fd);
#endif
	}
```

mtkaczyk (changes requested):
Hi Nigel,
I hope I helped you with this tricky reopen_mddev and manage_fork_fds.
Please see comment ( I don't test suggestions).

ncroxon:
Pushed a new version.

mtkaczyk:
I did little testing of `reopen_mddev()` to ensure that we are doing right thing:
```diff
diff --git a/util.c b/util.c
index 3809de00e844..a6f47e7fa19c 100644
--- a/util.c
+++ b/util.c
@@ -2387,12 +2387,15 @@ void reopen_mddev(int mdfd)
         * the same fd
         */
        char *devnm = fd2devnm(mdfd);
+
+       printf("mdfd flags: 0x%x\n", fcntl(mdfd, F_GETFL));
        int fd = open_dev(devnm);

        if (!is_fd_valid(fd))
                return;

        dup2(fd, mdfd);
+       printf("mdfd reopen flags: 0x%x\n", fcntl(mdfd, F_GETFL));

        close_fd(&fd);
 }
```
Unfortunately `O_CREAT` is not returned. For native r0d1 it returned:
```bash
./mdadm -I /dev/nvme4n1
mdfd flags: 0xc002
mdfd reopen flags: 0xc000
mdadm: /dev/nvme4n1 attached to /dev/md/vol0, which has been started.

```
it is same without the patch:
```
mdfd flags: 0xc002
mdfd reopen flags: 0xc000
mdadm: /dev/nvme4n1 attached to /dev/md/vol0, which has been started.
```
LGTM.

mtkaczyk (commented):
Hi @ncroxon,
I found 2 nits you have to resolve before I can merge it.

ncroxon:
Fixed issues....  pushed a new version

-Nigel

mtkaczyk (approved):

PR #58, opened 2024-07-18
mdadm: sysfs.c fix coverity issues

ncroxon:

* Event fixed_size_dest: You might overrun the 32-character fixed-size string "mdi->sys_name" by copying "devnm" without checking the length

* Event fixed_size_dest: You might overrun the 50-character fixed-size string "sra->text_version" by copying "buf + 9" without checking the length.

* Event string_overflow: You might overrun the 32-character destination string "dev->sys_name" by writing 256 characters from "de->d_name".

### DIFF
--- a/sysfs.c
+++ b/sysfs.c
@@ -139,7 +139,7 @@ int sysfs_init(struct mdinfo *mdi, int fd, char *devnm)
 		goto out;
 	if (!S_ISDIR(stb.st_mode))
 		goto out;
-	strcpy(mdi->sys_name, devnm);
+	strncpy(mdi->sys_name, devnm, sizeof(mdi->sys_name) - 1);
 
 	retval = 0;
 out:
@@ -179,6 +179,7 @@ struct mdinfo *sysfs_read(int fd, char *devnm, unsigned long options)
 			sra->array.major_version = -1;
 			sra->array.minor_version = -2;
 			strcpy(sra->text_version, buf+9);
+			sra->text_version[sizeof(sra->text_version) - 1] = '\0';
 		} else {
 			sscanf(buf, "%d.%d",
 			       &sra->array.major_version,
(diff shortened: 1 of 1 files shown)


sysfs.c
```diff
@@ -179,6 +179,7 @@ struct mdinfo *sysfs_read(int fd, char *devnm, unsigned long options)
 			sra->array.major_version = -1;
 			sra->array.minor_version = -2;
 			strcpy(sra->text_version, buf+9);
+			sra->text_version[sizeof(sra->text_version) - 1] = '\0';
```
mtkaczyk, line 182: why not?
```c
snprintf(sra->text_version, sizeof(sra->text_version), "/%s, buf + 9);
```

sysfs.c
```diff
@@ -340,6 +341,7 @@ struct mdinfo *sysfs_read(int fd, char *devnm, unsigned long options)
 
 		}
 		strcpy(dev->sys_name, de->d_name);
+		dev->sys_name[sizeof(dev->sys_name) - 1] = '\0';
```
mtkaczyk, line 344: same as above, can we use snprintf?

mtkaczyk (changes requested):
Hi Nigel,
To avoid updating last element please use snprintf.

sysfs.c
ncroxon: because, it produces an error....

sysfs.c: In function ‘sysfs_read’:
sysfs.c:181:81: error: ‘%s’ directive output may be truncated writing up to 4086 bytes into a region of size 50 [-Werror=format-truncation=]
  181 |                    snprintf(sra->text_version, sizeof(sra->text_version), "%s", buf + 9);

ncroxon: @mtkaczyk, this works...

-                       strcpy(sra->text_version, buf+9);
+                       char new_string[sizeof(sra->text_version)];
+                       strcpy(new_string, buf + 9);
+                       sprintf(sra->text_version, "%s", new_string);

mtkaczyk: ok, this way is fine then :)

mtkaczyk (approved):

mtkaczyk (approved):

PR #63, opened 2024-07-25
imsm: add read OROM form ACPI UEFI tables

bkucman:
OROM - IMSM hardware capabilities

EFI vars depends on userspace, they need to be mounted to be accessible. Sporadic problems have been observed with availability at an early assemble stage. It is not possible to fully synchronize EFI vars mounts with udev rules processing.

For above reason it is added read IMSM OROM from ACPI tables as an second way. This method will be used for SATA and VMD family controllers.

ACPI tables are generated by sysfs, earlier in the boot process, before the stage of RAID assembly. The way of loading OROM via EFI vars is retained, ACPI tables will be a backup way.

Two paths will be maintained, because IMSM hardware capabilities are necessary for RAID assembly during booting, so access to them must be provided.

V2: 
- fixes after Mariusz review
V3:
- fix incorrect GUID for RST_VMD_V

### DIFF
--- a/platform-intel.c
+++ b/platform-intel.c
@@ -577,6 +577,9 @@ static const struct imsm_orom *find_imsm_hba_orom(struct sys_dev *hba)
 
 #define SYS_EFI_VAR_PATH "/sys/firmware/efi/vars"
 #define SYS_EFIVARS_PATH "/sys/firmware/efi/efivars"
+#define ACPI_TABLES_PATH "/sys/firmware/acpi/tables/"
+#define ACPI_UEFI_TABLE_BASE_NAME "UEFI"
+#define ACPI_UEFI_DATA_OFFSET 52
 #define SCU_PROP "RstScuV"
 #define AHCI_PROP "RstSataV"
 #define AHCI_SSATA_PROP "RstsSatV"
@@ -584,10 +587,73 @@ static const struct imsm_orom *find_imsm_hba_orom(struct sys_dev *hba)
 #define VROC_VMD_PROP "RstUefiV"
 #define RST_VMD_PROP "RstVmdV"
 
-#define VENDOR_GUID \
+#define PCI_CLASS_RAID_CNTRL 0x010400
+
+/* GUID length in Bytes */
+#define GUID_LENGTH 16
+
+/* GUID entry in 'UEFI' for Sata controller. */
+#define RST_SATA_V_GUID \
+	EFI_GUID(0xe4dd92e0, 0xac7d, 0x11df, 0x94, 0xe2, 0x08, 0x00, 0x20, 0x0c, 0x9a, 0x66)
+
+/* GUID entry in 'UEFI' for sSata controller. */
+#define RST_SSATA_V_GUID \
+	EFI_GUID(0xb002be42, 0x901d, 0x4018, 0xb4, 0x1e, 0xd7, 0x04, 0xab, 0x3a, 0x0f, 0x15)
+
+/* GUID entry in 'UEFI' for tSata controller. */
+#define RST_TSATA_V_GUID \
+	EFI_GUID(0x101ce8f1, 0xb873, 0x4362, 0xa9, 0x76, 0xb5, 0x54, 0x31, 0x74, 0x52, 0x7e)
+
+/* GUID entry in 'UEFI' for Intel(R) VROC VMD. */
+#define RST_UEFI_V_GUID \
+	EFI_GUID(0x4bf2da96, 0xde6e, 0x4d8a, 0xa8, 0x8b, 0xb3, 0xd, 0x33, 0xf6, 0xf, 0x3e)
+
+/**
+ * GUID entry in 'UEFI' for Intel(R) RST VMD.
+ * Currently is the same like in 'UEFI' for Sata controller.
+ */
+#define RST_VMD_V_GUID RST_SATA_V_GUID
+
+/* GUID of intel RST vendor EFI var. */
+#define INTEL_RST_VENDOR_GUID \
 	EFI_GUID(0x193dfefa, 0xa445, 0x4302, 0x99, 0xd8, 0xef, 0x3a, 0xad, 0x1a, 0x04, 0xc6)
 
-#define PCI_CLASS_RAID_CNTRL 0x010400
+/*
+ * Unified Extensible Firmware Interface (UEFI) Specification Release 2.10
+ * UEFI ACPI DATA TABLE, Table O.1
+ */
+typedef struct uefi_acpi_table {
+	char signature[4];
+	__u32 length;
+	__u8 revision;
+	__u8 checksum;
+	char oemid[6];
+	/* controller name */
+	char oem_table_id[8];
+	__u32 oem_revision;
+	__u32 creator_id;
+	__u32 creator_revision;
+	/* controller GUID */
+	struct efi_guid identifier;
+	/* OROM data offeset */
+	__u16 dataOffset;
+} uefi_acpi_table_t;
+
+typedef struct uefi_acpi_table_with_orom {
+	struct uefi_acpi_table table;
+	struct imsm_orom orom;
+} uefi_acpi_table_with_orom_t;
+
+/* imsm_orom_id - Identifier used to match imsm efi var or acpi table
+ * @name: name of the UEFI property, it is part of efivar name or ACPI table oem_table_id
+ * @guid: acpi table guid identifier
+ *
+ * vendor guid (second part of evifar name) is not added here because it is cost.
+ */
+typedef struct imsm_orom_id {
+	char *name;
+	struct efi_guid guid;
+} imsm_orom_id_t;
 
 static int read_efi_var(void *buffer, ssize_t buf_size,
 			const char *variable_name, struct efi_guid guid)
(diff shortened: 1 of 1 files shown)


platform-intel.c
```diff
@@ -669,14 +724,251 @@ static int read_efi_variable(void *buffer, ssize_t buf_size,
...
+ * @guid: EFI guid.
+ * @guid1: EFI guid to compare.
+ *
+ * Return: %true if guid are equal, %false otherwise.
+ */
+bool is_efi_guid_equal(struct efi_guid guid, struct efi_guid guid1)
+{
+	int index;
+
+	for (index = 0; index < GUID_LENGTH; index++) {
```
mtkaczyk, line 738: GUID_LENGTH should be defined in `platform-intel.c` - it is used only here.

platform-intel.c
```diff
@@ -669,14 +724,251 @@ static int read_efi_variable(void *buffer, ssize_t buf_size,
...
+ * @uefi_controllers_number: number of UEFI controllers.
+ * @orom: OROM struct buffer to fill out.
+ *
+ * Find OROM in UEFI ACPI table based on requested controllers.
+ * If found, the buffer with the OROM structure will be filled.
+ *
+ * Return: %MDADM_STATUS_SUCCESS on success, %MDADM_STATUS_ERROR otherwise.
+ */
+mdadm_status_t
+find_orom_in_acpi_uefi_tables(uefi_controller_t *uefi_controllers, int uefi_controllers_number,
```
mtkaczyk, line 861: I miss explanation about unusual placement of it in ACPI tables. You have to clarify that it can be found in one of Intel provided `UEFI` ACPI table.

platform-intel.c
```diff
@@ -669,14 +724,251 @@ static int read_efi_variable(void *buffer, ssize_t buf_size,
...
+		if (!acpi_is_requested_uefi_controller(uefi_controllers, uefi_controllers_number,
+						       uefi_table.table))
+			continue;
+
+		if (!verify_uefi_acpi_table_length(uefi_table.table))
+			continue;
+
+		if (read_uefi_acpi_orom_data(fd, ent->d_name, &uefi_table)) {
+			continue;
+		} else {
```
mtkaczyk, line 904: you dint need `else` here.

platform-intel.c
```diff
@@ -669,14 +724,251 @@ static int read_efi_variable(void *buffer, ssize_t buf_size,
...
+			memcpy(orom, &uefi_table.orom, sizeof(uefi_table.orom));
+			status = MDADM_STATUS_SUCCESS;
+			goto close;
```
mtkaczyk, line 907: just `break`, we don't need `goto`

platform-intel.c
```diff
@@ -669,14 +724,251 @@ static int read_efi_variable(void *buffer, ssize_t buf_size,
...
+
+	for (index = 0; index < GUID_LENGTH; index++) {
+		if (guid.b[index] != guid1.b[index])
+			return false;
+	}
+	return true;
+}
+
+/**
+ * acpi_is_requested_uefi_controller() - check if read UEFI table contains requested controller.
```
mtkaczyk, line 746: we have to highlight in naming that we looking for one from the list (first match wins).

Passing array does not match the "controller" in the name of the function. Perhaps, you should move `for loop` up, and here just compare with the one passed:

```c
bool acpi_is_requested_uefi_controller(uefi_controller_t *uefi_controller, struct uefi_acpi_table table)
{
     if (strncmp(table.oem_table_id, uefi_controller->name,
		strlen(uefi_controller->name)) == 0)
        return false;
    return is_efi_guid_equal(table.identifier, uefi_controller->guid);
}
```

platform-intel.c
```diff
@@ -669,14 +724,251 @@ static int read_efi_variable(void *buffer, ssize_t buf_size,
...
+	closedir(dir);
+	return status;
+}
+
+/**
+ * find_imsm_efi_orom() - find OROM for requested controller.
+ * @orom: buffer for OROM.
+ * @controller_type: requested controller type.
+ *
+ * Based on controller type, function first search in EFI vars then in ACPI UEFI tables.
```
mtkaczyk, line 922: Please highlight that it is looking for one of the variables defined in arrays , i.e. first match wins.

platform-intel.c
```diff
@@ -669,14 +724,251 @@ static int read_efi_variable(void *buffer, ssize_t buf_size,
...
+	case SYS_DEV_SATA_VMD:
+		uefi_controllers = vmd_uefi_controllers;
+		uefi_controllers_number = ARRAY_SIZE(vmd_uefi_controllers);
+		break;
+	default:
+		return MDADM_STATUS_UNDEF;
+	}
+
+	/* Attempt to read via EFI vars. */
+	for (index = 0; index < uefi_controllers_number; index++)
```
mtkaczyk, line 958: For readability, please move this for to the function. It got me confused because once we have a loop and later we are using function. My confusion probably comes from first match wins conception used here.

```c
if (find_orom_in_efi_var_fs(uefi_controllers, uefi_controllers_number, orom);
	return MDADM_STATUS_SUCCESS;
return find_orom_in_acpi_uefi_tables(uefi_controllers, uefi_controllers_number, orom);
```

platform-intel.c
```diff
@@ -729,6 +1005,7 @@ const struct imsm_orom *find_imsm_efi(struct sys_dev *hba)
 	return &ret->orom;
 }
 
+
```
mtkaczyk, line 1008: This is unnecessary :)

mtkaczyk (commented):
This is good, few minor comments.

Thanks!

platform-intel.c
```diff
@@ -577,17 +577,72 @@ static const struct imsm_orom *find_imsm_hba_orom(struct sys_dev *hba)
...
+	/* OROM data offeset */
+	__u16 dataOffset;
+} uefi_acpi_table_t;
+
+typedef struct uefi_acpi_table_with_orom {
+	struct uefi_acpi_table table;
+	struct imsm_orom orom;
+} uefi_acpi_table_with_orom_t;
+
+typedef struct uefi_controller {
```
mtkaczyk, line 642: Please rename this to `imsm_orom_id` and add description.

this has nothing to do with controller which has different meaning in this code. orom is legacy name but it is still used as efi structure name, so it should be consistent.
Thanks!

platform-intel.c
```diff
@@ -577,17 +577,81 @@ static const struct imsm_orom *find_imsm_hba_orom(struct sys_dev *hba)
...
 #define VROC_VMD_PROP "RstUefiV"
 #define RST_VMD_PROP "RstVmdV"
 
-#define VENDOR_GUID \
+#define PCI_CLASS_RAID_CNTRL 0x010400
+
+/* GUID length in Bytes */
+#define GUID_LENGTH 16
+
+/* GUID entry in 'UEFI' for sSata controller. */
```
ktanska, line 595: Sata instead of sSata

platform-intel.c
```diff
@@ -577,17 +577,81 @@ static const struct imsm_orom *find_imsm_hba_orom(struct sys_dev *hba)
...
+#define RST_TSATA_V_GUID \
+	EFI_GUID(0x101ce8f1, 0xb873, 0x4362, 0xa9, 0x76, 0xb5, 0x54, 0x31, 0x74, 0x52, 0x7e)
+
+/* GUID entry in 'UEFI' for Intel(R) VROC VMD. */
+#define RST_UEFI_V_GUID \
+	EFI_GUID(0x4bf2da96, 0xde6e, 0x4d8a, 0xa8, 0x8b, 0xb3, 0xd, 0x33, 0xf6, 0xf, 0x3e)
+
+/* GUID entry in 'UEFI' for Intel(R) RST VMD. */
+#define RST_VMD_V_GUID \
+	EFI_GUID(0x4bf2da96, 0xde6e, 0x4d8a, 0xa8, 0x8b, 0xb3, 0xd, 0x33, 0xf6, 0xf, 0x3e)
```
ktanska, line 613: Just to make sure it is not copy-paste error - RST and VROC have the same guid? Where can I check it?

platform-intel.c
```diff
@@ -669,14 +733,274 @@ static int read_efi_variable(void *buffer, ssize_t buf_size,
...
+
+	dir = opendir(ACPI_TABLES_PATH);
+	if (!dir)
+		return MDADM_STATUS_ERROR;
+
+	for (ent = readdir(dir); ent; ent = readdir(dir)) {
+		close_fd(&fd);
+
+		/* Check if file is a UEFI table */
+		if (strncmp(ent->d_name, ACPI_UEFI_TABLE_BASE_NAME,
```
ktanska, line 889: What if there will be UEFI1/2/3 instead of UEFI?
-r-------- 1 root root     66 Aug  2 09:16 UEFI1
-r-------- 1 root root     92 Aug  2 09:16 UEFI2
-r-------- 1 root root     92 Aug  2 09:16 UEFI3
-r-------- 1 root root     92 Aug  2 09:16 UEFI4

bkucman: This check accepted all files starting with UEFI

ktanska: Ack, didn't see != 0 and was thinking == 0 is used :)

platform-intel.c
bkucman: yes, thanks, corrected

platform-intel.c
bkucman: done

platform-intel.c
ktanska: what about RST_VMD_V_GUID and RST_SATA_V_GUID - do they have the same guids?

bkucman: Yes

mtkaczyk (commented):
Few comments mainly related to over extended validation and error logging :) (I don't like to introduce dead code).

Functionally, it looks really good.

mtkaczyk (approved):

PR #64, opened 2024-07-25
imsm: refactor chunk size print

bkucman:
- add imsm_chunk_ops struct for better code readability,
- move chunk size mapping to string into array,
- add function to print supported chunk sizes by IMSM controller.


mku514k (approved):
LGTM

PR #66, opened 2024-07-30
mdmon: delegate removal to managemon

mtkaczyk:
Starting from [1], kernel requires suspend lock on member drive remove path. It causes deadlock with external management because monitor thread may be locked on suspend and is unable to switch array to active, for example if badblock is reported in this time.

It is blocking action now, so it must be delegated to managemon thread but we must ensure that monitor does metadata update first, just after detecting faulty.

This patch adds appropriative support. Monitor thread detects "faulty", and updates the metadata. After that it is asking manager thread to remove the device.

Manager should not activate recovery for the device if it is not able to remove all devices marked to remove.

Because it is blocking action in kernel now, looks like that retries are no longer required. There is no special short delay in case of remove failure as it was in monitor.

[1] kernel commit cfa078c8b80d ("md: use new apis to suspend array for adding/removing rdev from state_store()")

### DIFF
--- a/managemon.c
+++ b/managemon.c
@@ -440,6 +440,39 @@ static int disk_init_and_add(struct mdinfo *disk, struct mdinfo *clone,
 	return 0;
 }
 
+/**
+ * managemon_disk_remove()- remove disk from the MD array.
+ * @disk: device to be removed.
+ * @array_devnm: the name of the array to remove disk from.
+ *
+ * It tries to remove the disk from the MD array and if it is successful then it closes all opened
+ * descriptors. Removing action requires suspend, it might take a while.
+ * Invalidating mdi->state_fd will prevent from using this device further (see duplicate_aa()).
+ *
+ * To avoid deadlock, new file descriptor is opened because monitor may already wait on
+ * mdddev_suspend() in kernel and keep saved descriptor locked.
+ *
+ * Returns MDADM_STATUS_SUCCESS if disk has been removed, MDADM_STATUS_ERROR otherwise.
+ */
+static mdadm_status_t managemon_disk_remove(struct mdinfo *disk, char *array_devnm)
+{
+	int new_state_fd = sysfs_open2(array_devnm, disk->sys_name, "state");
+
+	if (!is_fd_valid(new_state_fd))
+		return MDADM_STATUS_ERROR;
+
+	if (write_attr("remove", new_state_fd) != MDADM_STATUS_SUCCESS)
+		return MDADM_STATUS_ERROR;
+
+	close_fd(&new_state_fd);
+	close_fd(&disk->state_fd);
+	close_fd(&disk->recovery_fd);
+	close_fd(&disk->bb_fd);
+	close_fd(&disk->ubb_fd);
+
+	return MDADM_STATUS_SUCCESS;
+}
+
 static void manage_member(struct mdstat_ent *mdstat,
 			  struct active_array *a)
 {
@@ -518,6 +551,43 @@ static void manage_member(struct mdstat_ent *mdstat,
 		if (write_attr("0.001", a->safe_mode_delay_fd) == MDADM_STATUS_SUCCESS)
 			a->info.safe_mode_delay = 1;
 
+	if (a->check_member_remove) {
+		bool any_removed = false;
+		bool all_removed = true;
+		struct mdinfo *disk;
+
+		for (disk = a->info.devs; disk; disk = disk->next) {
+			if (disk->man_disk_to_remove == false)
+				continue;
+
+			if (disk->mon_descriptors_not_used == false) {
+				/* To early, repeat later */
+				all_removed = false;
+				continue;
+			}
+
+			if (managemon_disk_remove(disk, a->info.sys_name)) {
+				all_removed = false;
+				continue;
+			}
+
+			any_removed = true;
+		}
+
+		if (any_removed) {
+			struct active_array *newa = duplicate_aa(a);
+
+			if (all_removed)
+				newa->check_member_remove = false;
+
+			replace_array(container, a, newa);
+			a = newa;
+		}
+
+		if (!all_removed)
+			return;
+	}
+
 	/* We don't check the array while any update is pending, as it
 	 * might container a change (such as a spare assignment) which
 	 * could affect our decisions.
@@ -539,8 +609,6 @@ static void manage_member(struct mdstat_ent *mdstat,
 			return;
 
 		newa = duplicate_aa(a);
-		if (!newa)
-			goto out;
 		/* prevent the kernel from activating the disk(s) before we
 		 * finish adding them
 		 */
@@ -570,7 +638,7 @@ static void manage_member(struct mdstat_ent *mdstat,
 				  "sync_action", "recover") == 0)
 			newa->prev_action = recover;
 		dprintf("recovery started on %s\n", a->info.sys_name);
- out:
+
 		while (newdev) {
 			d = newdev->next;
 			free(newdev);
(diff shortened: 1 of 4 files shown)


managemon.c
```diff
@@ -575,7 +634,7 @@ static void manage_member(struct mdstat_ent *mdstat,
 				  "sync_action", "recover") == 0)
 			newa->prev_action = recover;
 		dprintf("recovery started on %s\n", a->info.sys_name);
- out:
+
```
ktanska, line 637: new line not needed

ktanska (approved):
LGTM

bkucman (changes requested):

managemon.c
mtkaczyk: it is needed here, the next instruction is while.

monitor.c
```diff
@@ -621,24 +618,11 @@ static int read_and_act(struct active_array *a, fd_set *fds)
 			write_attr("-blocked", mdi->state_fd);
 		}
 
-		if ((mdi->next_state & DS_REMOVE) && mdi->state_fd >= 0) {
-			int remove_result;
-
-			/* The kernel may not be able to immediately remove the
-			 * disk.  In that case we wait a little while and
```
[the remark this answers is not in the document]
mtkaczyk: I removed this :)

bkucman (approved):

mtkaczyk:
WIP, we see some issues with it

mtkaczyk:
Getting it back, mdmon crash is fixed. It was caused by closing descriptors referred in `maxfd` in monitor.
We still observing regression in stress remove testing (removing and adding back same disk under IO multiple times). Working on this. We can move forward with review. Maybe I missed something..

mtkaczyk:
With the change, I'm observing that both threads are blocked in kernel during stress remove/plug testing:
managemon (expected):
```
==> /proc/776372/task/776372/stack <==
[<0>] mddev_suspend+0x17d/0x1e0 [md_mod]
[<0>] rdev_attr_store+0x11f/0x1e0 [md_mod]
[<0>] kernfs_fop_write_iter+0x128/0x1c0
[<0>] vfs_write+0x310/0x440
[<0>] ksys_write+0x5d/0xd0
[<0>] do_syscall_64+0x47/0x110
[<0>] entry_SYSCALL_64_after_hwframe+0x76/0x7e
```

monitor (not expected):
```
==> /proc/776372/task/776378/stack <==
[<0>] __fdget_pos+0x41/0x60
[<0>] ksys_write+0x2e/0xd0
[<0>] do_syscall_64+0x47/0x110
[<0>] entry_SYSCALL_64_after_hwframe+0x76/0x7e
```

If strace or gdb is beeing connected then monitor gets unblocked.  It requires further investigation. It might be connected to sharing the same resources.

mtkaczyk:
> With the change, I'm observing that both threads are blocked in kernel during stress remove/plug testing: managemon (expected):
> 
> ```
> ==> /proc/776372/task/776372/stack <==
> [<0>] mddev_suspend+0x17d/0x1e0 [md_mod]
> [<0>] rdev_attr_store+0x11f/0x1e0 [md_mod]
> [<0>] kernfs_fop_write_iter+0x128/0x1c0
> [<0>] vfs_write+0x310/0x440
> [<0>] ksys_write+0x5d/0xd0
> [<0>] do_syscall_64+0x47/0x110
> [<0>] entry_SYSCALL_64_after_hwframe+0x76/0x7e
> ```
> 
> monitor (not expected):
> 
> ```
> ==> /proc/776372/task/776378/stack <==
> [<0>] __fdget_pos+0x41/0x60
> [<0>] ksys_write+0x2e/0xd0
> [<0>] do_syscall_64+0x47/0x110
> [<0>] entry_SYSCALL_64_after_hwframe+0x76/0x7e
> ```
> 
> If strace or gdb is beeing connected then monitor gets unblocked. It requires further investigation. It might be connected to sharing the same resources.

Issue fixed. To avoid races like that, managemon is opening new file descriptor for remove because monitor might access `rdev/state` in the same time and hung because this fd is locked by managemon in kernel.

mdadm.h
```diff
@@ -408,8 +408,15 @@ struct mdinfo {
 	#define DS_BLOCKED	16
 	#define	DS_REMOVE	1024
 	#define	DS_UNBLOCK	2048
+	#define	DS_EXTERNAL_BB	4096
 	int prev_state, curr_state, next_state;
 
+	/* If set by monitor, managemon needs to remove faulty device */
+	bool man_disk_to_remove : 1;
```
mku514k, line 415: I think man_remove_disk would be sufficient, disk_to_remove sounds like you would specify a disk :D, but I's just me being nitpicky.

I'm fine either way.

monitor.c
```diff
@@ -147,14 +142,16 @@ int read_dev_state(int fd)
...
+		else if (sysfs_attr_match(cp, "write_mostly"))
 			rv |= DS_WRITE_MOSTLY;
-		if (sysfs_attr_match(cp, "spare"))
+		else if (sysfs_attr_match(cp, "spare"))
 			rv |= DS_SPARE;
-		if (sysfs_attr_match(cp, "blocked"))
+		else if (sysfs_attr_match(cp, "blocked"))
 			rv |= DS_BLOCKED;
+		else if (sysfs_attr_match(cp, "external_bbl"))
+			rv |= DS_EXTERNAL_BB;
```
mku514k, line 154: This could use maps if you'd like to make it extra legit, but I'm fine with this as it is.

mku514k (approved):

mdadm.h
mtkaczyk: Because I'm specifying it. At least I'm saying that this is a disk and this filed is for managemon.

mdinfo in abused in mdadm. I added mon, to let reader know what it the purpose of this filed.

mdinfo could be array or disk now :(

monitor.c
mtkaczyk: This should go separately because as you probably know, there will be more places like that :(
I think that I will revert this change to not mess with not important rework here.

mtkaczyk: And in fact, it was bad change.

managemon.c
```diff
@@ -440,6 +440,39 @@ static int disk_init_and_add(struct mdinfo *disk, struct mdinfo *clone,
...
+ * managemon_disk_remove()- remove disk from the MD array.
+ * @disk: device to be removed.
+ * @array_devnm: the name of the array to remove disk from.
+ *
+ * It tries to remove the disk from the MD array and if it is successful then it closes all opened
+ * descriptors. Removing action requires suspend, it might take a while.
+ * Invalidating mdi->state_fd will prevent from using this device further (see duplicate_aa()).
+ *
+ * To avoid deadlock, new file descriptor is opened because monitor may already wait on
+ * mdddev_suspend() in kernel and keep saved descriptor locked.
```
mtkaczyk, line 453: To avoid deadlock, new file descriptor is opened because monitor may need to read
state and this tread might be locked on suspends in kernel.

ktanska (approved):

monitor.c
```diff
@@ -643,17 +643,14 @@ static int read_and_act(struct active_array *a, fd_set *fds)
...
-		/* manager will do the actual check */
-		if (check_degraded)
-			a->check_degraded = 1;
-		if (check_reshape)
-			a->check_reshape = 1;
+	if (check_degraded || check_reshape || disks_to_remove) {
+
+		a->check_member_remove |= disks_to_remove;
+		a->check_degraded |= check_degraded;
+		a->check_reshape |= check_reshape;
```
mku514k, line 653: Can we move assignments outside the "if" statement and if any is set just call signal_manager()?

mku514k (approved):
LGTM

PR #68, opened 2024-07-31
Fix compilation error for gcc14

ktanska:
Fix compilation error. Set bit migr_state instead of assigning int
value.
Fix review.yml file, update gcc version used to build for each configuration from matrix.
Add new released gcc to compilation test during GH action.
Change runner to Ubuntu 24.04 which supports gcc versions up to 14.
Previously ubuntu-latest was used (22.04) which didn't support gcc 13
and 14. Add verification if correct gcc was installed during test.

### DIFF
--- a/super-intel.c
+++ b/super-intel.c
@@ -4324,6 +4324,12 @@ static void migrate(struct imsm_dev *dev, struct intel_super *super,
 static void end_migration(struct imsm_dev *dev, struct intel_super *super,
 			  __u8 map_state)
 {
+	/* To avoid compilation error, saying dev can't be NULL when
+	 * migr_state is assigned.
+	 */
+	if (dev == NULL)
+		return;
+
 	struct imsm_map *map = get_imsm_map(dev, MAP_0);
 	struct imsm_map *prev = get_imsm_map(dev, dev->vol.migr_state == 0 ?
 						    MAP_0 : MAP_1);
--- a/super-intel.c
+++ b/super-intel.c
@@ -194,6 +194,8 @@ ASSERT_SIZE(imsm_map, 52)
 struct imsm_vol {
 	__u32 curr_migr_unit_lo;
 	__u32 checkpoint_id;	/* id to access curr_migr_unit */
+#define MIGR_STATE_NORMAL 0
+#define MIGR_STATE_MIGRATING 1
 	__u8  migr_state;	/* Normal or Migrating */
 #define MIGR_INIT 0
 #define MIGR_REBUILD 1
@@ -4297,7 +4299,7 @@ static void migrate(struct imsm_dev *dev, struct intel_super *super,
 	struct imsm_map *dest;
 	struct imsm_map *src = get_imsm_map(dev, MAP_0);
 
-	dev->vol.migr_state = 1;
+	dev->vol.migr_state = MIGR_STATE_MIGRATING;
 	set_migr_type(dev, migr_type);
 	set_vol_curr_migr_unit(dev, 0);
 	dest = get_imsm_map(dev, MAP_1);
@@ -4331,7 +4333,7 @@ static void end_migration(struct imsm_dev *dev, struct intel_super *super,
 		return;
 
 	struct imsm_map *map = get_imsm_map(dev, MAP_0);
-	struct imsm_map *prev = get_imsm_map(dev, dev->vol.migr_state == 0 ?
+	struct imsm_map *prev = get_imsm_map(dev, dev->vol.migr_state == MIGR_STATE_NORMAL ?
 						    MAP_0 : MAP_1);
 	int i, j;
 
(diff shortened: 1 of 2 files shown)


super-intel.c
```diff
@@ -4357,7 +4357,7 @@ static void end_migration(struct imsm_dev *dev, struct intel_super *super,
 		map_state = imsm_check_degraded(super, dev, failed, MAP_0);
 	}
 
-	dev->vol.migr_state = 0;
+	dev->vol.migr_state |= 0;
```
mtkaczyk, line 4360: It was clearing it, now it is definitely not clearing it. I did simple program:
```c
#include <stdio.h>

int main()
{
    int v = 1;
    printf("%d\n", v);
    
    v |= 0;
    
    printf("%d\n", v);

    return 0;
}
```
Result is: 
>1
>1

This is wrong to me.

mtkaczyk (changes requested):

gene-git:
- I note that the compile error goes away if I use "-O1" instead of "-O2".

 - Compiler is also perfectly happy (including -O2)  with assigning to zero if I make a small file using the struct definitions taken from super-intel.c and a simple function:

```
// struct definitions here
int main(int argc, char *argv[]) 
{   
    struct imsm_dev idev ;  
    struct imsm_dev *dev = &idev;
    
    dev->vol.migr_state = 0;
    printf(" 1. %d\n", dev->vol.migr_state);
    return (0);
}
```

So something about the larger super-intel.c goes awry with -O2.

mtkaczyk:
@gene-git thanks for feedback! it should be helpful.

In fact, I don't understand where is the issue and why this is a fix (Kinga did just a refactor to use defines) :(

@ktanska I need a detailed description in commit message because I don't understand why changing `0` to `#define MIGR_NORMAL 0` is fixing our issue. There must be something we are missing.

bkucman:
> @gene-git thanks for feedback! it should be helpful.
> 
> In fact, I don't understand where is the issue and why this is a fix (Kinga did just a refactor to use defines) :(
> 
> @ktanska I need a detailed description in commit message because I don't understand why changing `0` to `#define MIGR_NORMAL 0` is fixing our issue. There must be something we are missing.

It is probably false positive issue of gcc: https://stackoverflow.com/questions/75184243/wstringop-overflow-compiler-warning-in-custom-fmtformatterstdbitsetn-sp

gene-git:
I may be missing something, but I tried to  replacing '0' with a macro which did not help anything - same warning. 
i.e. around line 4366:

```
#define MIGR_NORMAL 0
    dev->vol.migr_state = MIGR_NORMAL;
```
Leads to:
```foo.c: In function ‘end_migration’:
foo.c:4366:29: error: writing 2 bytes into a region of size 0 [-Werror=stringop-overflow=]
 4366 |         dev->vol.migr_state = MIGR_NORMAL;
```

It (obviously) compiles fine with the warning suppressed ( -Wno-stringop-overflow )

mtkaczyk:
@ktanska we have to disable `stringop-overflow` because it it obviously buggy.

There is no possibility that the size is 0 so it must be false positive.

gene-git:
I have found a possible workaround -  this keeps compiler happy.

Simply add (or move) the offending line above the loop. I am not familiar with the code but this looks like it should not change anything at all - right? The dev->vol.migr_state is set unconditionally to 0 after the loop and a quick read suggests it is not used / changed inside the loop. 

i.e. At line 4343:
```
    dev->vol.migr_state = 0;
    if (map_state != map->map_state && (is_gen_migration(dev) == false) &&
```

Compiles without the warning/error. 

Maybe this is helpful.

peluse (changes requested):
Agree, this is a little odd and I'm not sure why it is not failing with just the #define change.  I would strongly suggest changing the workflow with a different PR though in case that is masking the original failure somehow (I didn't play with it so could be wrong, just suggesting)

gene-git:
@peluse 
> ...not sure why it is not failing with just the #define change. 

Fyi just in case you missed some comments above.  As noted in comments here :
  - In my tests a #define makes no difference at all and compile fails the same way
  -There is a  proposed fix which does work:   add "if (dev && dev->vol.migr_state != 0)" at line 4366

mtkaczyk:
> I have found a workaround - this keeps compiler happy.
> 
> Change line 4366 to be conditional:
> 
> ```
>  if (dev && dev->vol.migr_state != 0)
>         dev->vol.migr_state = 0;
> ```
> 
> Compiles without the warning/error. It also has the benefit of checking the dev pointer.
> 
> This allows us to keep the stringop-overflow turned on.

What about:

```c
assert(dev);
```

at the beginning of `end_migration` function?
I believe it is working but this code has limited sense in this place. Probably we don't need to check migr_state value, we just need to ensure that it is not NULL, that is why I propose `assert()`.

mtkaczyk:
@gene-git please confirm that this fixes issues on your side!

ktanska:
> I've decided to use assert(), because it seems the most clear solution and resolves issue reported by compiler.

And one more small fix from me - use if/return instead of assert - assert can be off with -DNDEBUG flag.

gene-git:
> 
> And one more small fix from me - use if/return instead of assert - assert can be off with -DNDEBUG flag.

I confirm this works fine for me as well. 
Thank you all for sorting this (tricky little issue) out.

mtkaczyk (approved):

PR #71, opened 2024-08-04
md.4: replace wrong word

nicoroeser:
There is a wrong word in the `md(4)` man page; the commit in this pull request corrects it.

### DIFF
--- a/md.4
+++ b/md.4
@@ -224,7 +224,7 @@ option.  If you use this option to
 while running a newer kernel, the array will NOT assemble, but the
 metadata will be update so that it can be assembled on an older kernel.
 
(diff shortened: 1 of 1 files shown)


mtkaczyk:
Hey @nicoroeser
Thanks for this, fell free to do more detailed review if you want (for sure there are many different errors) :)

Please add commit message to satisfy our review GitHub job, because we are obligated to follow kernel coding style (copy, pull description there but don't use markdown).

Thanks!

nicoroeser:
Here you are, @mtkaczyk, the updated commit should solve the problem.

About more detailed reviews: of course I’ll send more patches in case I spot additional problems.

Thanks for your work on the software!

PR #73, opened 2024-08-08
imsm: get bus from VMD driver directory

mtkaczyk:
Enumeration od VMD child devices is started early, kernel is not waiting for VMD enumeration to finish. It causes that:
`/sys/bus/pci/drivers/vmd/{dev}/domain/device` link might be not ready yet.

With PCI gen5 devices we can observe that mdadm is failing to start IMSM raid arrays because of that. In that case, it needs to find bus path manually.

Look for bus device in VMD driver directory if realpath failed with ENOENT.

### DIFF
--- a/platform-intel.c
+++ b/platform-intel.c
@@ -105,12 +105,75 @@ static void free_sys_dev(struct sys_dev **list)
 	}
 }
 
+/**
+ * vmd_find_pci_bus() - look for PCI bus created by VMD.
+ * @vmd_path: path to vmd driver.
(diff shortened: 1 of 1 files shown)


bkucman (approved):
Hi,

LGTM 

Just two typos to fix.
First in commit desc
`Enumeration **od** VMD child devices is ` > `Enumeration **of** VMD child devices is

bkucman (approved):

mku514k (approved):
LGTM

PR #78, opened 2024-08-26
Introduce new way of assemblation (v2)

mku514k:
This patchset changes current way of assemblation, so external metadata volumes are assembled but not started unitll sufficient number of drives is incrementally added.  This means that not started volumes will now be visible in mdstat. This is basically what https://github.com/md-raid-utilities/mdadm/commit/4dde420fc3e24077ab926f79674eaae1b71de10b tried to achieve but it compromised on many things. Until array is safe to start, only necessary sysfs entries are written to make array visible in mdstat. To force writing sysfs with full array info `--force-set-array` option is introduced.

About last-resort: This changes allow for last-resort feature to work for external metadata arrays. We will handle this by making it configurable (probably via conf file) in the future, in separate patchset.

### DIFF
--- a/Assemble.c
+++ b/Assemble.c
@@ -1983,9 +1983,10 @@ int assemble_container_content(struct supertype *st, int mdfd,
 	int old_raid_disks;
 	int start_reshape;
 	char *avail;
-	int err;
-	int is_clean, all_disks;
-	bool is_raid456;
+	int err = 0;
+	int is_clean, all_disks, missing_disks;
+	bool is_raid456, need_rebuild;
+	char buf[SYSFS_MAX_BUF_SIZE] = {0};
 
 	if (sysfs_init(content, mdfd, NULL)) {
 		pr_err("Unable to initialize sysfs\n");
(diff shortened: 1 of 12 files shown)


mku514k:
Checkpatch returns an error due to leading spaces, but this is how the rest of the lines are. Can we live with this check not passing or you want me to replace spaces with tab for every `long_options[]` line?

mtkaczyk:
> Checkpatch returns an error due to leading spaces, but this is how the rest of the lines are. Can we live with this check not passing or you want me to replace spaces with tab for every `long_options[]` line?

That is fine.  What you can eventually do is to propose separate PR with the change and then we can rebase this one.

mtkaczyk:
This one will be merged after new release.

mtkaczyk:
This one is risky and AFAIC Intel is leaving VROC business therefore taking it is not justified from user experience point of view as there might not be a support with issues. It is working current way more than 10 years, issues it is solving are not critical, just better handling of "failed/broken" arrays.

Sorry @mku514k! You did a great job here but Intel changed course..

PR #79, opened 2024-08-27
mdadm: Remove limit of 127 in md device name intorudced in 25aa732.

shmsingh:
Hi, A limit was introduced in [2511732](https://github.com/md-raid-utilities/mdadm/commit/25aa7329141c0b28d8811671627f0f5c5dc22273) which is causing failure if the number in device name is > 127. This patch is removing that limit.


mtkaczyk:
> Remove limit of 127 in md device name intorudced in https://github.com/md-raid-utilities/mdadm/commit/25aa7329141c0b28d8811671627f0f5c5dc22273.

This is interesting that checkpatch passed here, @ktanska could you please take a look?
I expected fail, this should be: commit <12+SHA>("<TITLE>")

Going back to the change. I added a limit on purpose, I don't think that someone needs more 127 arrays. I think that people are using it for grouping arrays. I limited it from security reasons (and maybe I wrongly assumed that it is a limit in kernel, I don't remember now).

For sure I cannot accept just dropping it (user input must be always validated). You can extend the limit if you need to, but there is a better way for grouping arrays (and in the future I would like to enable it as a primary way). It is called "named arrays"
You can enabled it by setting "CREATE names=yes" in conf.

Thanks,
Mariusz

shmsingh:
> > This is interesting that checkpatch passed here, @ktanska could you please take a look?
> > I expected fail, this should be: commit <12+SHA>("<TITLE>")
> 
> Checkpatch will not fail on this problem. Without writing "Commit" before its SHA or add tag "Fixes", checkpatch treats this SHA like a hyperlink. @shmsingh please rewrite commit message and add "Fixes" tag.

Updated the commit message.

shmsingh:
> > Remove limit of 127 in md device name intorudced in [25aa732](https://github.com/md-raid-utilities/mdadm/commit/25aa7329141c0b28d8811671627f0f5c5dc22273).
> 
> This is interesting that checkpatch passed here, @ktanska could you please take a look? I expected fail, this should be: commit <12+SHA>("<TITLE>")
> 
> Going back to the change. I added a limit on purpose, I don't think that someone needs more 127 arrays. I think that people are using it for grouping arrays. I limited it from security reasons (and maybe I wrongly assumed that it is a limit in kernel, I don't remember now).
> 
> For sure I cannot accept just dropping it (user input must be always validated). You can extend the limit if you need to, but there is a better way for grouping arrays (and in the future I would like to enable it as a primary way). It is called "named arrays" You can enabled it by setting "CREATE names=yes" in conf.
> 
> Thanks, Mariusz

Hi Marisuz, 

I uderstand your point, but adding this limit is breaking the existing automation framework. Switching the framework to use named array will take some time. Its not like we have 127+ devices , but it's good to have the flexibility to use number greater than 127.

mtkaczyk:
Ok, that fine. Please then replace `127` by the number that satisfied your needs and is acceptable by kernel but please keep the limit as low as possible.

Please also mention this in commit description, that it broke your testing suite. It is regression for sure and it was intended (I decided to keep it low) so any change in this area must be justified by real use case now.

Now, we are adjusting hard requirement to the real user needs.

shmsingh:
> Ok, that fine. Please then replace `127` by the number that satisfied your needs and is acceptable by kernel but please keep the limit as low as possible.
> 
> Please also mention this in commit description, that it broke your testing suite. It is regression for sure and it was intended (I decided to keep it low) so any change in this area must be justified by real use case now.
> 
> Now, we are adjusting hard requirement to the real user needs.

After discussing internally we decided to go with 1024 as the limit. This limit covers the numbers used in exisiting framework and also has enough room left for any future test cases.

mtkaczyk (approved):

PR #80, opened 2024-08-28
mdadm: Change displaying of devices in --detail

asztukow:
The counts of active, working, failed and spare devices were not printed when the number was zero.

Refactor the code to always display the counts of all device types, regardless of their number. This way, it is more reliable for users.


ktanska (approved):

mtkaczyk:
I agree. I see hiding these information as less clear for users.
We should exclude them only if the type of member device is not supported by type of mddevice (for example active_disks for external container). We can limit it in the future if we will see need to.
For now, print it loud!

PR #86, opened 2024-09-03
CI: use self-hosted runner to run tests

ktanska:
Use prepared VM machine in GitHub actions to run mdadm tests on it.

### DIFF
--- a/.github/tools/run_mdadm_tests.sh
+++ b/.github/tools/run_mdadm_tests.sh
@@ -1,0 +1,10 @@
+#!/usr/bin/bash
+
+sudo make clean
+sudo make -j$(nproc)
+sudo make install
+sudo mdadm -Ss
+sudo ./test setup
+ret=$(sudo ./test --skip-broken --no-error --disable-integrity --disable-multipath --disable-linear --keep-going)
+sudo ./test cleanup
+exit $ret
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,58 @@
+name: tests
+on:
+  schedule:
+    - cron: "0 0 * * *"
+  pull_request:
+    paths:
+      - '*.c'
+      - '*.h'
+jobs:
+  upstream_tests:
+     if: ${{ github.repository == 'md-raid-utilities/mdadm' }}
+     runs-on: self-hosted
+     timeout-minutes: 150
+     name: upstream tests
+     steps:
+     - uses: actions/checkout@v4
+       if: ${{ github.event_name == 'pull_request' }}
+       with:
+        ref: ${{ github.event.pull_request.head.sha }}
+        fetch-depth: 0
+     - uses: actions/checkout@v4
+       if: ${{ github.event_name == 'schedule' }}
+       with:
+        ref: main
+        fetch-depth: 0
+     - name: 'Prepare machine'
+       run: |
+          cd ..
+          vagrant halt
+          vagrant status
+          vagrant up
+     - name: 'Run tests'
+       run: |
+          cd ..
+          vagrant ssh -c "cd /home/vagrant/host/mdadm && .github/tools/run_mdadm_tests.sh"
+     - name: 'Copy logs to host machine'
+       run: |
+          cd ..
+          vagrant ssh -c "sudo mkdir -p /home/vagrant/host/logs && sudo mv /var/tmp/*.log /home/vagrant/host/logs"
+     - name: "Save artifacts"
+       uses: actions/upload-artifact@v4
+       with:
+         name: "Logs from failed tests"
+         path: /home/ci/actions-runner/_work/mdadm/logs/*.log
+     - name: "Clean logs"
+       run: |
+          cd ..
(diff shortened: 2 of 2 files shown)


.github/workflows/review.yml
```diff
@@ -43,3 +43,26 @@ jobs:
       run: mv .github/tools/.checkpatch.conf .
     - name: Run checkpatch review
       uses: webispy/checkpatch-action@v9
+  upstream_tests:
+     runs-on: self-hosted
```
peluse, line 47: take a look at the condition in the workflow of mdraid that make's sure the repo is exactly this one before using the self-hosted runner, we don't want forks using our runners.

.github/workflows/review.yml
```diff
@@ -43,3 +43,26 @@ jobs:
...
+     timeout-minutes: 120
```
peluse, line 48: 120 minutes?  Is that accurate?  What is the average test time?

.github/workflows/review.yml
```diff
@@ -43,3 +43,26 @@ jobs:
...
+     name: upstream tests
+     steps:
+     - uses: actions/checkout@v4
+       with:
+        ref: ${{ github.event.pull_request.head.sha }}
+        fetch-depth: 0
+     - name: 'Prepare machine'
+       run: |
+          vagrant snapshot restore clean_vm_v1
```
peluse, line 57: although I believe this works I think you need to add "cd .." here (unless its above somewhere and I missed it.  it's best to run vagrant commands in the dir where Vagrantfile lives and by default that is one level up from where each step starts on the runner

.github/workflows/review.yml
```diff
@@ -43,3 +43,26 @@ jobs:
...
+          echo "FYI vagrant restore command finished with exit code: $?"
+          sleep 2
+          vagrant status
```
peluse, line 60: cool, FYI that these (sleep and status) aren't "required" but in bringing up mdraid I found them helpful to deal with some flakiness with vagrant every once in a while and its nice in the worfklow output to confirm the status of the box every time you bring it up

.github/workflows/review.yml
```diff
@@ -43,3 +43,26 @@ jobs:
...
+          vagrant up
+          sleep 2
+     - name: 'Run tests'
+       run: |
+          vagrant ssh -c "cd host/mdadm && .github/tools/run_mdadm_tests.sh"
```
peluse, line 65: same comment here, every step should start with "cd ..".  I do think it works without that but I think it's safest to be in the exact dir as we'll add more runners to this same sever later after you get things working with one.  They are isolated by dirs

.github/workflows/review.yml
ktanska: I changed it to the lower value - 90 minutes, and unfortunately tests on this machine takes more than 1 hour. Discussed with Mariusz and he is OK with waiting for the results. 
It depends on platform, the fastest run which I saw was 30 minutes.

ktanska:
Logs from run could be find here: https://github.com/md-raid-utilities/mdadm/actions/runs/11738905324/job/32702330296?pr=86 
Tests didn't start for this pr, because they only start when commit contains changes in .c or .h files.

.github/tools/run_mdadm_tests.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/bash
+
+sudo make clean
+sudo make -j$(nproc)
+sudo make install
+sudo mdadm -Ss
+sudo ./test setup
+sudo ./test --skip-broken --no-error --disable-integrity --disable-multipath --disable-linear --keep-going
+sudo ./test cleanup
```
mtkaczyk, line 9: Please add emplty line at the end.

.github/tools/run_mdadm_tests.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/bash
+
+sudo make clean
+sudo make -j$(nproc)
+sudo make install
+sudo mdadm -Ss
+sudo ./test setup
+sudo ./test --skip-broken --no-error --disable-integrity --disable-multipath --disable-linear --keep-going
```
mtkaczyk, line 8: we should provide test status up:
```suggestion
ret=$(sudo ./test --skip-broken --no-error --disable-integrity --disable-multipath --disable-linear --keep-going)
sudo ./test cleanup
exit(ret)
```

.github/workflows/nightly_tests.yml
```diff
@@ -0,0 +1,47 @@
+name: nightly_tests
+on:
+  schedule:
+    - cron: "0 0 * * *"
+jobs:
+  upstream_tests:
+     if: ${{ github.repository == 'md-raid-utilities/mdadm' }}
+     runs-on: self-hosted
+     timeout-minutes: 120
```
mtkaczyk, line 9: Shoudn't it be aligned to same value as in `tests.yml`? if it executes same test suite, can we have execution in one yml to avoid such differences?

mtkaczyk (commented):
LGTM but for consistency I would like to keep `run_tests` shared code in one yml file (if possible).

mtkaczyk (approved):

PR #87, opened 2024-09-03
Incremental: support devnode in IncrementalRemove.

mtkaczyk:
There are no reasons to keep this interface different than others. Allow to use devnode but keep old way for backward compatibility. Method is added to verify that only devnode or kernel name is used.

Testing on my setup passed. I used raid0, so removal failed, but I checked parsing.
```bash
[gklab-localhost ~/mtkaczyk_work/mdadm/mdadm-github]# ./mdadm -If /dev/nvme5n1
mdadm: set device faulty failed for nvme5n1:  Device or resource busy
mdadm: nvme5n1 is still in use, cannot remove.
[gklab-localhost ~/mtkaczyk_work/mdadm/mdadm-github]# ./mdadm -If /dev/md/nvme5n1
mdadm: Cannot remove "/dev/md/nvme5n1", devnode path or kernel device name is allowed.
[gklab-localhost ~/mtkaczyk_work/mdadm/mdadm-github]# ./mdadm -If /dev/disk/by-path/pci-0000\:00\:11.5-ata-5-part4
mdadm: Cannot remove "/dev/disk/by-path/pci-0000:00:11.5-ata-5-part4", devnode path or kernel device name is allowed.
[gklab-localhost ~/mtkaczyk_work/mdadm/mdadm-github]# ./mdadm -If nvmex5n1

```

### DIFF
--- a/Incremental.c
+++ b/Incremental.c
@@ -1674,33 +1674,52 @@ static void remove_from_member_array(struct mdstat_ent *memb,
 	}
 }
 
-/*
- * IncrementalRemove - Attempt to see if the passed in device belongs to any
- * raid arrays, and if so first fail (if needed) and then remove the device.
+/**
+ * is_devnode_path() - check if the devname passed might be devnode path.
+ * @devnode: the path to check.
  *
- * @devname - The device we want to remove
- * @id_path - name as found in /dev/disk/by-path for this device
+ * Devnode must be located directly in /dev directory. It is not checking existence of the file
+ * because the device might no longer exist during removal from raid array.
+ */
+static bool is_devnode_path(char *devnode)
+{
+	char *devnm = strrchr(devnode, '/');
+
+	if (!devnm || *(devnm + 1) == 0)
+		return false;
+
+	if (strncmp(devnode, DEV_DIR, DEV_DIR_LEN) == 0 && devnode + DEV_DIR_LEN - 1 == devnm)
+		return true;
+
+	return false;
+}
+
+/**
+ * IncrementalRemove() - Remove the device from all raid arrays.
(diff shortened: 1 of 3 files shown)


Incremental.c
```diff
@@ -1674,49 +1674,68 @@ static void remove_from_member_array(struct mdstat_ent *memb,
...
+static bool is_devnode_path(char *devnode)
+{
+	char *devnm = strrchr(devnode, '/');
+
+	printf("%p %p\n", devnode + DEV_DIR_LEN, devnm);
+
+	if (devnm && strncmp(devnode, DEV_DIR, DEV_DIR_LEN) == 0
+	    && devnode + DEV_DIR_LEN - 1 == devnm)
+		return true;
+
```
ktanska, line 1693: How about "/dev/" - I've checked and it returns true, is it OK?

ktanska (approved):

Incremental.c
```diff
@@ -1674,49 +1674,68 @@ static void remove_from_member_array(struct mdstat_ent *memb,
...
-/*
- * IncrementalRemove - Attempt to see if the passed in device belongs to any
- * raid arrays, and if so first fail (if needed) and then remove the device.
+/**
+ * is_devnode() - check if the devname passed might be devnode path.
+ * @devnode: the path to check.
  *
- * @devname - The device we want to remove
- * @id_path - name as found in /dev/disk/by-path for this device
+ * Devnode must be placed directly in /dev directory. It is not checking existence of the file
```
mku514k, line 1681: I'm being picky here, but consider using "located" instead of "placed", is not something we have control of.

...also there should be no "s" in "exists".

mku514k (approved):
LGTM, only small nitpicks from me.

PR #96, opened 2024-09-25
Refactor: Add sysfs_write_descriptor()

mtkaczyk:
Function `IncrementalRemove` is changed to `Incremental_remove`, following convention`<file>_<name>`. Less important, not dangerous change.

Function `sysfs_write_descriptor` is added to have some unification in handling errors for write(). The main motivation is to ensure that `errno` is copied immediately and is reliable. 
Also, verification varies in many places: (ret == -1 or ret == (strlen(buf)), errno might be not reliable is all scenarios.


sysfs.c
```diff
@@ -73,6 +73,46 @@ void sysfs_free(struct mdinfo *sra)
...
+
+/**
+ * sysfs_write_descriptor()- wrapper for write(), projected to be used with sysfs.
+ * @fd: file descriptor.
+ * @value: value to set.
+ * @len: length of the value.
+ * @errno_p: On write() failure, buffer to copy errno value, might be NULL.
+ *
+ * Errors are differentiated, because (at least theoretically) kernel may not process whole string,
+ * It is different type of problem, so decision belongs to caller. @err is not updated in this case
```
ktanska, line 96: kernel may not process whole string, which is a different type of problem, so decision belongs to caller.

ktanska: Looks like sentence was not finished

mtkaczyk: Updated.

ktanska (approved):

mku514k (approved):
LGTM

Incremental.c
```diff
@@ -1695,15 +1695,15 @@ static bool is_devnode_path(char *devnode)
...
+ * Incremental_remove() - Remove the device from all raid arrays.
  * @devname: the device we want to remove, it could be kernel device name or devnode.
  * @id_path: optional, /dev/disk/by-path path to save for bare scenarios support.
  * @verbose: verbose flag.
  *
  * First, fail the device (if needed) and then remove the device from native raid array or external
  * container.  If it is external container, the device is removed from each subarray first.
  */
-int IncrementalRemove(char *devname, char *id_path, int verbose)
+int Incremental_remove(char *devname, char *id_path, int verbose)
```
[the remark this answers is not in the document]
mtkaczyk: Sure, in case of libraries it is preferred to start function with the file name, I followed this concept for public function. 
There is static lib target, probably dead now but I may be restored in any moment.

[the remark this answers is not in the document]
mku514k: Ack

PR #97, opened 2024-09-25
mdadm: add xmalloc.h

mtkaczyk:
Move memory declaration helpers outside mdadm.h. They seems to be useful so keep them but include separatelly. Rework them to not reffer to Name[] declared internally in mdadm/mdmon.

This is first step to start decomplexing mdadm.h.

### DIFF
--- a/Examine.c
+++ b/Examine.c
@@ -22,14 +22,16 @@
  *    Email: <neilb@suse.de>
  */
 
-#include	"mdadm.h"
 #include	"dlink.h"
+#include	"mdadm.h"
+#include	"md_u.h"
+#include	"md_p.h"
+#include	"xmalloc.h"
 
 #if ! defined(__BIG_ENDIAN) && ! defined(__LITTLE_ENDIAN)
 #error no endian defined
 #endif
-#include	"md_u.h"
-#include	"md_p.h"
+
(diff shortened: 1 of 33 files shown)


Examine.c
```diff
@@ -23,6 +23,7 @@
  */
 
 #include	"mdadm.h"
+#include	"xmalloc.h"
```
ktanska, line 26: new line missing

mtkaczyk:
SPDX style is adjusted to other files as recommended by Linux Foundation:
https://spdx.dev/learn/handling-license-info/

xmalloc.c
```diff
@@ -21,64 +21,55 @@
...
+
+#include <string.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+
+static void *exit_memory_alloc_failure(void)
+{
+	fprintf(stderr, "Memory allocation failure - aborting\n");
+	exit(5);
```
ktanska, line 34: Why 5? Maybe we can consider exit with 1 for failure?

mku514k: I wonder about that too. If this is for legacy reasons, I'd appreciate a comment why this specific value is used.

mku514k (approved):

PR #98, opened 2024-09-25
Mdmonitor: Fix startup with missing directory

asztukow:
Commit a07dea8d3b78 ("Mdmonitor: Refactor check_one_sharer() for better error handling") introduced an issue, if directory /run/mdadm is missing, monitor fails to start. Move the directory creation earlier to ensure it is always created.

### DIFF
--- a/Monitor.c
+++ b/Monitor.c
@@ -225,6 +225,11 @@ int Monitor(struct mddev_dev *devlist,
 		return 1;
 	}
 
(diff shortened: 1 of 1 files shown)


mtkaczyk (approved):

ktanska (approved):

mku514k (changes requested):
I know this is a nitpick, but you missed first number from commit hash, that's why checkpatch complains, it should be 0a07dea8d3b78 (with 0 at the beginning).

mku514k (approved):

PR #107, opened 2024-10-07
Man update

mtkaczyk:
As requested in #105 I reviewed all references to legacy kernels in the code and removed them.

Fixes: #105 

Additionally, I realized that `auto={yes, no,mdp,partitions}` is probably dead. I also tried to remove it.


mtkaczyk:
@neilbrown could you please confirm that support for `auto=` can be removed?
It looks safe but you never know, maybe I'm missing something..

mku514k:
I see small inconsistency. "Auto" is being removed alongside documentation, yet we silently ignore this option usage. I get that this is for compatibility reasons, but shouldn't we at least print deprecation warning so people who still use it will stop doing so?
Imo, if we want to completely get rid of all "auto" mentions in the future, this is the way.

mtkaczyk:
> I see small inconsistency. "Auto" is being removed alongside documentation, yet we silently ignore this option usage. I get that this is for compatibility reasons, but shouldn't we at least print deprecation warning so people who still use it will stop doing so?
Imo, if we want to completely get rid of all "auto" mentions in the future, this is the way.

Yup, info message is added now.

ktanska (approved):

PR #109, opened 2024-10-09
util: use only /dev directory in open_dev()

ktanska:
Previously, open_dev() tried to open device in two ways - using /dev and /tmp directory. This method could be used by users which have no access to /tmp directory (e.g. udev) and dev_open() fails which may affect many processes. Remove try to open in /tmp directory.

### DIFF
--- a/util.c
+++ b/util.c
@@ -1111,17 +1111,6 @@ int dev_open(char *dev, int flags)
 			fd = open(devname, flags);
 			unlink(devname);
 		}
-		if (fd < 0) {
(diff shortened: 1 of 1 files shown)


mtkaczyk:
I think it is safe change. It is more probable that `/tmp` is read only rather than `/dev` (especially for systemd fake users)
If mdadm process is failing because of it, modes for the user which is executing `mdadm` should be adjusted.

mtkaczyk (approved):

PR #110, opened 2024-10-10
Fix 07reshape5initr test 

ktanska:
This test could hang if "check" action is not written to sync_action. If
this value didn't appear, test hanged on infinite while loop. Add 5
second timeout to loop.

### DIFF
--- a/tests/07reshape5intr
+++ b/tests/07reshape5intr
@@ -31,7 +31,21 @@ do
   echo 1000 > /proc/sys/dev/raid/speed_limit_min
   echo 2000 > /proc/sys/dev/raid/speed_limit_max
   check wait
-  while ! echo check > /sys/block/md0/md/sync_action; do sleep 0.1; done
+
+  max=5
+
+  for ((i = 0 ; i < max ; i++ )); do
+    if echo check > /sys/block/md0/md/sync_action; then
+        break
+    fi
+    sleep 1
+  done
+
+  if [ i == max ]; then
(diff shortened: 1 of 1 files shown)


mtkaczyk:
This must wait till #109 is merged, or you can rebase it now.

ktanska:
@mtkaczyk change was rebased.

tests/07reshape5intr
```diff
@@ -31,7 +31,19 @@ do
   echo 1000 > /proc/sys/dev/raid/speed_limit_min
   echo 2000 > /proc/sys/dev/raid/speed_limit_max
   check wait
-  while ! echo check > /sys/block/md0/md/sync_action; do sleep 0.1; done
+
+  counter=0
+  max_count=50  # 50 iterations of 0.1s = 5 seconds
+ 
+  while ! echo check > /sys/block/md0/md/sync_action; do
```
mtkaczyk, line 38: Please use for or rework because exit condition should not be checked inside while (code readability reasons)

```
for ((i = 0 ; i < max ; i++ )); do
    if [[ echo check > /sys/block/md0/md/sync_action ]]; then
        break;
done

if [[ i == max ]]; then
     echo >&2 "Timeout waiting for check to succeed"
     exit 1
fi
```

mtkaczyk (changes requested):

ktanska:
Review will be continued in #113

PR #111, opened 2024-10-10
Start reshape after switchroot

mku514k:
There are numerous issues for --grow --continue in switchroot phrase, they include:
- Events being missed for restarting grow-continue service. This is apparent mostly on OS on RAID scenarios. When a checkpoint (next step) is commited, we have no reliable way to gracefully stop reshape until it reaches that checkpoint. During boot, there's heavy I/O utilisation, which causes sync speed drop, and naturally checkpoint takes longer to reach. This further causes systemd to forcefully kill grow-continue service due to timeouts, which results in udev event being missed for grow-continue service restart.
- Grow-continue (seemingly) was not designed to be restarted without reassembly, some things like stopping chunksize (to lower) migration were straight up not working until recently.

This patchset makes grow-continue (actual reshape) start after switchroot phrase. This way we should not encounter issues related to restarting the service. Major changes include:
- Added reshape status to detail export;
- Replaced current udev rule for grow continue, to be based on newly added entry in detail export;
- Added check not to start grow-continue if in initrd, to rely on udev trigger after switchroot;
- Added hints/prints for restoring the array to usable state if rehsape was not properly stopped.


Grow.c
```diff
@@ -3068,6 +3069,39 @@ static void catch_term(int sig)
...
+	/* Just fork if no service. */
+	if (service) {
+		/* Allow udev to trigger service after switchroot. */
+		if (in_initrd())
+			return 1;
+		if (continue_via_systemd(devname, GROW_SERVICE, NULL))
+			return 1;
+	}
+
+	switch (fork()) {
```
mtkaczyk, line 3098: why not just `return fork()`

Grow.c
```diff
@@ -3068,6 +3069,39 @@ static void catch_term(int sig)
...
+ *           0 if currently child process,
+ *           1 process forked, passed to service or delegated to udev.
+ */
+static int handle_forking(bool forked, char *devname, bool service)
+{
+	/* Nothing to do if already forked. */
+	if (forked)
+		return 0;
+
+	/* Just fork if no service. */
```
mtkaczyk, line 3089: this comment is not true, or maybe variable name in not detailed enough. Probably it should be `use_systemd`

Service could be there, we don't want it

Grow.c
```diff
@@ -3068,6 +3069,39 @@ static void catch_term(int sig)
...
+	if (service) {
+		/* Allow udev to trigger service after switchroot. */
```
mtkaczyk, line 3091: This is allowing anything, this is blocking service from beeing started in initramfs. that is all.

As a result of that, reshape can be always successfully started after switch root.

mtkaczyk: it requires additional explanation why only systemd way is affected.
Maybe instead bool `use_systemd` we should check if systemd is configured on this system and adjust the behavior for both ways.

If there is no systemd, then I would like to have similar logic for forking path.

Grow.c
```diff
@@ -3761,23 +3786,20 @@ int reshape_container(char *container, char *devname,
 	 */
 	ping_monitor(container);
 
-	if (!forked && !c->freeze_reshape)
-		if (continue_via_systemd(container, GROW_SERVICE, NULL))
-			return 0;
-
-	switch (forked ? 0 : fork()) {
+	/* Do not start a service if reshape frozen. */
+	switch (handle_forking(forked, container, !c->freeze_reshape)) {
```
mtkaczyk, line 3790: `!c->freeze_reshape` is totally not understandable now.

we cannot use systemd but we may still fork. Please unify those ways.

Grow.c
mtkaczyk: If you are concerned about `who` will start switch root later in the case without systemd- it is not your problem. It is problem of system manager.

They can check details and call grow-continue if needed.

util.c
```diff
@@ -2355,6 +2354,7 @@ int continue_via_systemd(char *devnm, char *service_name, char *prefix)
 	int pid, status;
 	char pathbuf[1024];
 
+	dprintf("Kickstart %s!\n", service_name);
```
mtkaczyk, line 2357: Please remove it or make it more understandable.

mtkaczyk (commented):
I'm fine with this if you will prove that, delaying reshape start is not blocking for array to be accessible (all array data can be read- nothing is suspended by default)

Grow.c
mku514k: Just for unification reasons, so we don't return 1s or PIDs depending what path is taken.

mku514k:
I decided to get rid of the helper / common function for handling forking, I don't have an idea how to do it properly.

Grow.c
```diff
@@ -3761,7 +3764,8 @@ int reshape_container(char *container, char *devname,
 	 */
 	ping_monitor(container);
 
-	if (!forked && !c->freeze_reshape)
```
mku514k, line 3764: @mtkaczyk The check was added in 40b941b813e22b33ff9c70831415f5b65e0fd70b

It states:

> Grow: Do not fork via systemd if freeze_reshape is set
> 
> Mdadm should not run 'grow-continue' unit file for container if
> '--freeze-reshape' argument is passed. Otherwise it will be ignored,
> and reshape will start anyway.

mku514k: Looks like this check might not be needed rn if we rely on just being in initrd. A helper function now should be easy to implement. Let me know what do you think bout it.

Grow.c
```diff
@@ -3068,12 +3069,40 @@ static void catch_term(int sig)
...
+ * Returns: -1 if fork() failed,
+ *           0 if child process,
+ *           1 if job delegated to forked process or systemd.
+ *
+ * This function is a helper function for reshapes for fork handling.
+ */
+static int handle_forking(bool forked, char *devname)
+{
+	if (forked)
+		return 0;
```
mtkaczyk, line 3087: Probably you should add that if `forked` is set it assumes that we are in child process already, right?

Grow.c
mtkaczyk: Yes, LGTM.

Grow.c
```diff
@@ -3566,33 +3597,24 @@ static int reshape_array(char *container, int fd, char *devname,
...
-			sysfs_free(sra);
-			return 0;
-		}
-
 	/* Now we just need to kick off the reshape and watch, while
 	 * handling backups of the data...
 	 * This is all done by a forked background process.
 	 */
-	switch(forked ? 0 : fork()) {
+	switch (handle_forking(forked, container ?: sra->sys_name)) {
```
mtkaczyk, line 3604: Please check if there is something that can be passed here for native (if that is a case).

mtkaczyk: if not, probably some assert that for native it must be forked, would be helpful. I'm not sure here.

mku514k:
Note to myself: Need to remove "--freeze-reshape" from man.

Grow.c
mku514k: Looks like rn we rely on service start to fail for native, and default to forking. Cursed.
image

mku514k: I can't think of scenario for native in which code reaches here, nor scenario for imsm in which sys_name will be used. 
Note that `container ?: sra->sys_name` will pass `container` if it's not null.

@mtkaczyk I would opt for leaving this as it is. This was "working" all these years when "--freeze-reshape" was a thing.
This already is quite dangerous change, can we deal with all "container" thing later?

mku514k: Discussed offline

mku514k:
I did some basic tests and it looks to me there's no problem with `--grow --continue` service based reshapes with said service not being active. There seems to be no blocked sectors for r/w.

Having said that, I have to have a deeper look to see if this code is really working. I tried messing with `--freeze-reshape` and even tough the flag was passed, the service still has been triggered. Maybe udev has something to do with it...

mku514k:
> Having said that, I have to have a deeper look to see if this code is really working. I tried messing with `--freeze-reshape` and even tough the flag was passed, the service still has been triggered. Maybe udev has something to do with it...

Ok, I retested this with my changes, and verified there should be no issues with udev triggering service before switchroot. The subtle change is that `--freeze-reshape` was working only for incremental and assembly, it would not work when passed with `--grow --continue`. Right now, due to the fact that we check if we're in initrd for reshapes globally, the `--grow --continue` service will respect it too; it'll get triggered but will quit.

On the other hand, I noticed with these changes there's quite a delay (~2s) between calling `incremental` for the last drive in the array and it actually executing. That could cause potential race/timeout issues. Gotta look deeper why this happens.

mku514k:
Looks to me as if we've got mdmon race... meh...
image

In terms of incremental, sometimes it waits at select() in recv_buf(), other times it finishes and then when one mdmon finally kills another the reshape is visible.

mku514k:
@mtkaczyk I added one commit on top for refactoring `continue_via_systemd()` as we discussed.

mtkaczyk:
Nice clean up, looks (almost) safe.
Merging.

PR #112, opened 2024-10-22
mdadm: Grow.c distinguish takeover from reshape on grow operation

ncroxon:
Correcting the terminology on the output when doing a takeover vs a reshape.

### DIFF
--- a/Grow.c
+++ b/Grow.c
@@ -2036,7 +2036,8 @@ int Grow_reshape(char *devname, int fd,
 		sysfs_free(sra);
 		return 1;
 	} else if (frozen < 0) {
-		pr_err("%s is performing resync/recovery and cannot be reshaped\n", devname);
+		pr_err("%s is performing resync/recovery and cannot be %s\n", devname,
+		       (s->level != UnSet && s->level != array.level) ? "taken over" : "reshaped");
 		sysfs_free(sra);
 		return 1;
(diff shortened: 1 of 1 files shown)


Grow.c
```diff
@@ -2036,7 +2036,8 @@ int Grow_reshape(char *devname, int fd,
 		sysfs_free(sra);
 		return 1;
 	} else if (frozen < 0) {
-		pr_err("%s is performing resync/recovery and cannot be reshaped\n", devname);
+		pr_err("%s is performing resync/recovery and cannot be %s\n", devname,
+		       (s->level != UnSet && s->level != array.level) ? "taken over" : "reshaped");
```
mtkaczyk, line 2040: Well, in cmdline we are setting new `--level` not `--take-over` and that is why I see it not clear (you need to know that level change == take over.)

What about?
```c
		pr_err("%s is performing resync/recovery and it cannot %s\n", devname,
 			(s->level != UnSet && s->level != array.level) ? "change level" : "be reshaped");`
```

mauelsha: Ok with me, as "takeover" is not covered in the MD userspace documentation and neither addressed on the mdadm command line as you mentioned , in fact "LEVEL CHANGES" are documented  in mdadm(4).
Thanks,
Heinz

PR #113, opened 2024-10-24
func.sh: do not hang when grow-continue can't finish

ktanska:
When grow-continue process is on, the sync_action represents value which means that recovery is in progress. When grow-continue does not finish, even if sync_action is not reshape anymore, test should fail.

### DIFF
--- a/tests/07reshape5intr
+++ b/tests/07reshape5intr
@@ -31,7 +31,21 @@ do
   echo 1000 > /proc/sys/dev/raid/speed_limit_min
   echo 2000 > /proc/sys/dev/raid/speed_limit_max
   check wait
-  while ! echo check > /sys/block/md0/md/sync_action; do sleep 0.1; done
+
+  max=5
+
+  for ((i = 0 ; i < max ; i++ )); do
+    if [[ $(echo check > /sys/block/md0/md/sync_action) != 0 ]]; then
+        break;
+    fi
+    sleep 1
+  done
+
+  if [[ i == max ]]; then
+     echo >&2 "Timeout waiting for check to succeed"
+     exit 1
+  fi
+
   check wait
   mm=`cat /sys/block/md0/md/mismatch_cnt`
   if [ $mm -gt 0 ]
--- a/tests/func.sh
+++ b/tests/func.sh
@@ -202,6 +202,24 @@ restore_selinux() {
 	setenforce $sys_selinux
 }
 
+wait_for_reshape_end() {
+	# wait for grow-continue to finish but break if sync_action does not
+	# contain any reshape value
+	while true
+	do
+		sync_action=$(grep -Ec '(resync|recovery|reshape|check|repair) *=' /proc/mdstat)
+		if (( "$sync_action" != 0 )); then
+			sleep 1
+			continue
+		elif [[ $(pgrep -f "mdadm --grow --continue" > /dev/null) != "" ]]; then
+			echo "Grow continue did not finish but reshape is done" >&2
+			exit 1
+		else
+			break
+		fi
+	done
+}
+
 setup_systemd_env() {
 	warn "Warning! Test suite will set up systemd environment!\n"
 	echo "Use \"systemctl show-environment\" to show systemd environment variables"
@@ -357,15 +375,28 @@ check() {
 		max=`cat /proc/sys/dev/raid/speed_limit_max`
 		echo 200000 > /proc/sys/dev/raid/speed_limit_max
 		sleep 0.1
-		while grep -Eq '(resync|recovery|reshape|check|repair) *=' /proc/mdstat ||
-			grep -v idle > /dev/null /sys/block/md*/md/sync_action
+		iterations=0
+		# Wait 10 seconds for one of the actions appears in sync_action.
+		while [ $iterations -le 10 ]
 		do
-			sleep 0.5
+			sync_action=$(grep -Ec '(resync|recovery|reshape|check|repair) *=' /proc/mdstat)
+			if (( "$sync_action" == 0 )); then
+				sleep 1
+				iterations=$(( $iterations + 1 ))
+				continue
+			else
+				break
+			fi
 		done
-		while ps auxf | grep "mdadm --grow --continue" | grep -v grep
-		do
-			sleep 1
-		done
+		echo "Reshape has not started after 10 seconds"
+
+		# Now let's wait for reshape to finish.
+		echo "Waiting for grow-continue to finish"
+		wait_for_reshape_end
+		# If we have matrix-raid there's a second process ongoing
+		sleep 5
+		wait_for_reshape_end
+
 		echo $min > /proc/sys/dev/raid/speed_limit_min
 		echo $max > /proc/sys/dev/raid/speed_limit_max
 	;;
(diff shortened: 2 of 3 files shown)


tests/07reshape5intr
```diff
@@ -31,7 +31,21 @@ do
   echo 1000 > /proc/sys/dev/raid/speed_limit_min
   echo 2000 > /proc/sys/dev/raid/speed_limit_max
   check wait
-  while ! echo check > /sys/block/md0/md/sync_action; do sleep 0.1; done
+
+  max=5
+
+  for ((i = 0 ; i < max ; i++ )); do
+    if [[ echo check > /sys/block/md0/md/sync_action ]]; then
```
mku514k, line 38: Does this statement work?
I believe bash will try to interpret this as conditional statement and fail, it would interpret it as a literal and command won't be run. I believe we can run the command directly without the square brackets.

tests/07reshape5intr
```diff
@@ -31,7 +31,21 @@ do
...
+        break;
```
mku514k, line 39: ";" not needed

tests/07reshape5intr
```diff
@@ -31,7 +31,21 @@ do
...
+    fi
+    sleep 1
+  done
+
+  if [[ i == max ]]; then
```
mku514k, line 44: You can use casual brackets here for numeric expressions.

tests/func.sh
```diff
@@ -357,14 +357,33 @@ check() {
...
+		# which will trigger grow-continue process
+		for ((i = 0 ; i < max_iterations ; i++ )); do
+			sync_action=`grep -Eq '(resync|recovery|reshape|check|repair) *=' /proc/mdstat || grep -v idle > /dev/null /sys/block/md*/md/sync_action`
+			if [ "$sync_action" ]; then
+				break
+			fi
+			sleep 1
 		done
+
+		if [[ i == max_iterations ]]; then
```
mku514k, line 371: Please use causal brackets for numeric comparisons.

tests/func.sh
```diff
@@ -357,14 +357,33 @@ check() {
...
+			echo >&2 "Timeout waiting for reshape operation appeared in sync_action"
+			exit 1
+		fi
+
+		# wait for grow-continue to finish but break if sync_action does not
+		# contain any reshape value
 		while ps auxf | grep "mdadm --grow --continue" | grep -v grep
```
mku514k, line 378: We could possibly use pgrep here like
`while pgrep -f "mdadm --grow --continue" > /dev/null`
...but I'm fine with leaving this as it is now.

tests/func.sh
```diff
@@ -357,14 +357,33 @@ check() {
...
 		sleep 0.1
-		while grep -Eq '(resync|recovery|reshape|check|repair) *=' /proc/mdstat ||
-			grep -v idle > /dev/null /sys/block/md*/md/sync_action
-		do
-			sleep 0.5
+		max_iterations=5
+		# wait up to 5 seconds for one of action appeared in sync_action,
+		# which will trigger grow-continue process
+		for ((i = 0 ; i < max_iterations ; i++ )); do
+			sync_action=`grep -Eq '(resync|recovery|reshape|check|repair) *=' /proc/mdstat || grep -v idle > /dev/null /sys/block/md*/md/sync_action`
```
mku514k, line 364: Second grep will be executed only if first one fails, is this what we want?

tests/func.sh
```diff
@@ -357,14 +357,33 @@ check() {
 		max=`cat /proc/sys/dev/raid/speed_limit_max`
 		echo 200000 > /proc/sys/dev/raid/speed_limit_max
 		sleep 0.1
-		while grep -Eq '(resync|recovery|reshape|check|repair) *=' /proc/mdstat ||
-			grep -v idle > /dev/null /sys/block/md*/md/sync_action
-		do
-			sleep 0.5
+		max_iterations=5
+		# wait up to 5 seconds for one of action appeared in sync_action,
+		# which will trigger grow-continue process
```
mku514k, line 362: Consider something like this
> Wait up to 5 seconds for one of actions to appear in sync_action.
> This should trigger grow-continue process.

tests/func.sh
```diff
@@ -357,14 +357,33 @@ check() {
...
+		# contain any reshape value
 		while ps auxf | grep "mdadm --grow --continue" | grep -v grep
 		do
-			sleep 1
+			sync_action=`grep -Eq '(resync|recovery|reshape|check|repair) *=' /proc/mdstat || grep -v idle > /dev/null /sys/block/md*/md/sync_action`
+			if [ "$sync_action" ]; then
+				sleep 1
+				continue
+			fi
+			echo >&2 "Grow continue does not finish but reshape is done"
```
mku514k, line 385: Consider "did not finish".

mku514k (changes requested):
Few requests from me.

tests/func.sh
ktanska: It is compatible with the original implementation, I think it should be done this way

mku514k:
@ktanska I see you marked the conversation as resolved, but I do not see any changes pushed here for the first patch.

Do you plan to update first commit?

tests/07reshape5intr
mtkaczyk, line 38: Please use direct comparison, I understand what you are doing but I would like to keep it as understandable as possible.
```suggestion
if [[ $(echo check > /sys/block/md0/md/sync_action) != 0 ]]; then
```

mtkaczyk (approved):
LGTM

tests/func.sh
```diff
@@ -357,14 +357,35 @@ check() {
...
+				((max_iterations--))
+			else
+				break
+			fi
 		done
-		while ps auxf | grep "mdadm --grow --continue" | grep -v grep
+
+		# wait for grow-continue to finish but break if sync_action does not
+		# contain any reshape value
+		while pgrep -f "mdadm --grow --continue" > /dev/null
```
mtkaczyk, line 375: Please move this to separate function because the block is not readable to me.

Caller should tell us that we should verify reshape instead of optimistically looping for it. Maybe it is not an issue now, but direct parameter is better readable (you always know when this code is executed).

mtkaczyk: Now it depends on your knowledge in the raid area, I want to lower the complexity of jumping into development.

mtkaczyk (changes requested):
Some comments for patch 2.

mtkaczyk (approved):
much better

ktanska:
@mtkaczyk @mku514k please review
Logs from run with this pr you can find here: https://github.com/md-raid-utilities/mdadm/actions/runs/11738905324/job/32702330296?pr=86

tests/func.sh
```diff
@@ -202,6 +202,24 @@ restore_selinux() {
...
+	# wait for grow-continue to finish but break if sync_action does not
+	# contain any reshape value
+	while true
+	do
+		sync_action=$(grep -Ec '(resync|recovery|reshape|check|repair) *=' /proc/mdstat)
+		if (( "$sync_action" != 0 )); then
+			sleep 1
+			continue
+		elif [[ $(pgrep -f "mdadm --grow --continue" > /dev/null) != "" ]]; then
+			echo >&2 "Grow continue did not finish but reshape is done"
```
mtkaczyk, line 215:
```suggestion
echo "Grow continue did not finish but reshape is done" >&2
```

mtkaczyk (approved):

PR #114, opened 2024-10-25
[PATCH] Coverity check resource fixes

ncroxon:
Converity is flagging the following errors:y
Event check_return: Calling without checking return value. This library function may fail and return an error code.

### DIFF
--- a/managemon.c
+++ b/managemon.c
@@ -387,14 +387,17 @@ static void manage_container(struct mdstat_ent *mdstat,
 
 static int sysfs_open2(char *devnum, char *name, char *attr)
 {
+	char buf[SYSFS_MAX_BUF_SIZE];
+
 	int fd = sysfs_open(devnum, name, attr);
 	if (fd >= 0) {
 		/* seq_file in the kernel allocates buffer space
 		 * on the first read.  Do that now so 'monitor'
 		 * never needs too.
 		 */
-		char buf[200];
-		if (read(fd, buf, sizeof(buf)) < 0)
+		int n = read(fd, buf, sizeof(buf));
+
+		if (n <= 0 || n == sizeof(buf))
 			/* pretend not to ignore return value */
 			return fd;
 	}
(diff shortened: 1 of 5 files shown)


managemon.c
```diff
@@ -387,14 +387,17 @@ static void manage_container(struct mdstat_ent *mdstat,
...
 	if (fd >= 0) {
 		/* seq_file in the kernel allocates buffer space
 		 * on the first read.  Do that now so 'monitor'
 		 * never needs too.
 		 */
-		char buf[200];
-		if (read(fd, buf, sizeof(buf)) < 0)
+		int n = read(fd, buf, sizeof(buf));
+
+		if (n <= 0 || n == sizeof(buf))
```
mtkaczyk, line 400: These checks are pointless (even in original version). we are pretending that it matters, but we are not really interested in what is returned, we are fine either way. The changes you proposed are only making it worse :(

Please understand the comment above and determine if it is still a case, maybe we can remove this code. I need you to do investigation here.

monitor.c
```diff
@@ -485,7 +485,10 @@ static int read_and_act(struct active_array *a, fd_set *fds)
 		 * wasn't requested, transition to read-auto.
 		 */
 		char buf[64];
-		read_attr(buf, sizeof(buf), a->metadata_fd);
+		int n = read_attr(buf, sizeof(buf), a->metadata_fd);
+
+		if (n <= 0)
```
mtkaczyk, line 490: shouldn't we act like it should be readonly if we cannot read property? I would not stop mdmon in case of we cannot read the property.

I hope it is only theoretical problem..

mtkaczyk (changes requested):
```
Converity is flagging the following errors:y
Event check_return: Calling without checking return value.
This library function may fail and return an error code.

Signed-off-by: Nigel Croxon <ncroxon@redhat.com>
```
Typo Coverity, please remove ":y"
Please also see comments, this one is not simple.

Thanks!

PR #115, opened 2024-10-25
Coverity fixes resources leaks

ncroxon:
Handle variable going out of scope leaks the handle.


mtkaczyk (changes requested):

mtkaczyk (approved):
I will merge it after the release.

PR #118, opened 2024-11-06
mdopen: remove wrong condition

mtkaczyk:
After mentioned patch, this condition get opposite meaning and it is blocking creation in cases where it was supported.

Remove it now.

Fixes: 119cdcad049e ("mdadm: drop auto= support")


ktanska (approved):
LGTM
Tested 'mdadm -Ss && mdadm -As' with and without patch and now it works :)
Creation works on my setup too.

bkucman (approved):

PR #120, opened 2024-11-07
imsm: add print license for VMD

bkucman:
Add print IMSM license for VMD controllers in --detail-platform. The license specifies the scope of RAID support in the platform for the VMD controller.


super-intel.c
```diff
@@ -645,6 +645,14 @@ struct imsm_update_rwh_policy {
 	int dev_idx;
 };
 
+enum imsm_sku {
+	SKU_NO_KEY = 0,
+	SKU_STANDARD_KEY,
```
mtkaczyk, line 650: In this case enum values are meaningful, please keep assignments for every enum value, do not let compiler or developer to guess.

mtkaczyk (changes requested):
Please set each enum value for extra security. I want to force compiler to set them always right.

PR #127, opened 2024-11-22
imsm: fix tpv drvies check in add_to_super

bkucman:
Before the mentioned patch, the check to verify if IMSM on current platform supports a use of TPV (other than Intel) disk, was only performed for non-Intel disks, after it is performed for all. This change causes inability to use any disk when platform does not support TPV drives, attempt results in the following error.

mdadm:  Platform configuration does not support non-Intel NVMe drives.
        Please refer to Intel(R) RSTe/VROC user guide.

This change restores the check if the disk is non-Intel.

Fixes: 734e7db4dfc5 ("imsm: Remove warning and refactor add_to_super_imsm code")


super-intel.c
```diff
@@ -6122,7 +6122,8 @@ static int add_to_super_imsm(struct supertype *st, mdu_disk_info_t *dk,
 			pr_err("%s controller supports Multi-Path I/O, Intel (R) VROC does not support multipathing\n",
 			       basename(cntrl_path));
 
-		if (super->orom && !imsm_orom_has_tpv_support(super->orom)) {
+		if (devpath_to_vendor(pci_dev_path) != 0x8086 && super->orom &&
```
mtkaczyk, line 6125: super->orom should come first. I think compiler will realize that but I don't need to consider compilation if params are passed in right order, so please change it:
```suggestion
		if (super->orom && devpath_to_vendor(pci_dev_path) != 0x8086
```

mtkaczyk (changes requested):
one nit, please reverse order.

super-intel.c
bkucman: done

mtkaczyk (approved):

PR #130, opened 2024-11-26
mdadm/Manage: Don't write superblock when slave nodes cluster confirm…

t460p:
… in MD cluster

A clustered disk is added by the traditional --add sequence. However, other nodes may not  "see" the device by "blkid -d device -t devie_uuid=xxxx".


t460p:
need commit again

PR #134, opened 2024-12-02
Fix tests to make the CI clear

mku514k:
Fix few tests to clear the CI.

### DIFF
--- a/clustermd_tests/01r10_Grow_bitmap-switch
+++ b/clustermd_tests/01r10_Grow_bitmap-switch
@@ -19,7 +19,7 @@ check all nobitmap
 mdadm --grow $md0 --bitmap=internal
 [ $? -eq '0' ] ||
 	die "$NODE1: change bitmap 'none' to 'internal' failed."
-sleep 1
+sleep 2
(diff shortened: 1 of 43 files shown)


ncroxon:
The job running on runner inspur5 has exceeded the maximum execution time of 150 minutes.

mtkaczyk:
@ncroxon 
This is separate problem. At this moment we focused on making CI happen. Mateusz is trying to solve many issues in tests to make them consistently passing. From other hand, we need to take a look into test suite and decide:
- which tests are valuable to be run for mdadm? (here we are the most focused on testing mdadm functionalities)
- what we can do to make to execution time shorter?

I added logging test time to know which tests are the most time consuming but I don't have time to analyze that yet, sorry :( any help is welcomed.

The goal is to make tests job passing to have clear results for PRs (tests job fail = regression caused by PR).

mtkaczyk:
We don't have a much time to make it happen so superficially LGTM.

What I miss here is a another filter option to start marking mdadm regression tests and kernel regression tests (kernel regression tests are not needed to be run for mdadm).

This probably needs to be addressed in separate pull, maybe @ncroxon would you like to challenge it?

PR #142, opened 2025-01-18
Including missing xmalloc header in raid6check to fix FTBFS.

daniel-baumann:
Hi,

with current default Debian build-flags, mdadm doesn't compile anymore because there's a missing header in raid6check.c. I've applied this in Debian some time ago and it would be nice if you could merge it.

Regards,
Daniel

### DIFF
--- a/raid6check.c
+++ b/raid6check.c
@@ -23,6 +23,8 @@
  */
 
 #include "mdadm.h"
+#include "xmalloc.h"
+
 #include <stdint.h>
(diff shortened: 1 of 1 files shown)


daniel-baumann:
...oh, and jftr the additional empty line is just to make it consistent with the other code.

daniel-baumann:
Hi,

thanks for the pointer - updated and pushed.

Regards,
Daniel

PR #143, opened 2025-01-23
udev: persist properties of MD devices after switch_root

aafeijoo-suse:
dracut installs in the initrd a custom udev rule only to set this option [1]. The main purpose is that if an MD device is activated in the initrd, its properties are kept on the udev database after the transition from the initrd to the rootfs. This was added to fix detection issues when LVM is on top.

This patch would allow to remove the custom udev rule shipped by dracut (`63-md-raid-arrays.rules` is already being installed in the initrd), and it will also benefit other initrd generators that do not want to create custom udev rules.

[1] https://github.com/dracutdevs/dracut/blob/5d2bda46f4e75e85445ee4d3bd3f68bf966287b9/modules.d/90mdraid/59-persistent-storage-md.rules#L23

### DIFF
--- a/udev-md-raid-arrays.rules
+++ b/udev-md-raid-arrays.rules
@@ -29,6 +29,7 @@ ENV{DEVTYPE}=="partition", ENV{MD_DEVNAME}=="*[0-9]", SYMLINK+="md/$env{MD_DEVNA
 IMPORT{builtin}="blkid"
 OPTIONS+="link_priority=100"
 OPTIONS+="watch"
+OPTIONS+="db_persist"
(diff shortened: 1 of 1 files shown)


mtkaczyk (approved):
LGTM

aafeijoo-suse:
> @aafeijoo-suse Can you please clean up this error: WARNING:COMMIT_LOG_LONG_LINE: Possible unwrapped commit description (prefer a maximum 75 chars per line)

Sorry, I'm not sure how to reference the URL in [1] using only 75 chars. Should I remove it from the commit message?

mtkaczyk:
> Sorry, I'm not sure how to reference the URL in [1] using only 75 chars. Should I remove it from the commit message?

No, that is fine. In this case it is unavoidable, `checkpatch` is not perfect. I prefer to have reference.

PR #145, opened 2025-01-27
mdadm: fix --grow with --add for linear

mtkaczyk:
For the case mdadm --grow with --add, the s.btype should not be initialized yet, hence BitmapUnknown should be checked instead of BitmapNone.

Noted that this behaviour should only support by md-linear, which is removed from kernel, howerver, it turns out md-linear is used widely in home NAS and we're planning to reintroduce it soon.

Fixes: 581ba1341017 ("mdadm: remove bitmap file support")

PR to test Kuai's changes.


ncroxon:
trying to figure out why the testing failed.

PR #149, opened 2025-02-03
mdadm: use <linux/raid> headers

mtkaczyk:
For a years we redefined these headers in mdadm. It was efficient way to handle implementation of new features, we didn't rely on kernel during compilation. It is huge mess now.

We should reuse the headers exported to integrate driver and mdadm better.  Add them and remove local headers.

This PR requires deep review and testing. @ncroxon, @XiaoNi87 I need your support on this.

### DIFF
--- a/Create.c
+++ b/Create.c
@@ -23,8 +23,6 @@
  */
 
 #include	"mdadm.h"
-#include	"md_u.h"
-#include	"md_p.h"
 #include	"udev.h"
 #include	"xmalloc.h"
 
--- a/mdadm.h
+++ b/mdadm.h
@@ -34,6 +34,7 @@ extern __off64_t lseek64 __P ((int __fd, __off64_t __offset, int __whence));
 #endif
 
 #include	<assert.h>
+#include	<asm/byteorder.h>
 #include	<sys/types.h>
 #include	<sys/stat.h>
 #include	<stdarg.h>
@@ -85,7 +86,6 @@ struct dlm_lksb {
 #endif
 
 #include	<linux/kdev_t.h>
-/*#include	<linux/fs.h> */
 #include	<sys/mount.h>
 #include	<asm/types.h>
 #include	<sys/ioctl.h>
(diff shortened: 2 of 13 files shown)


mdadm.c
ncroxon, line 1: I'm good with this change.

mdadm.h
ncroxon, line 1: I'm good with this change... It does require the super1.c change in this patch set to compile.

mdmonitor.c
ncroxon, line 1: I'm good with this change.

mdopen.c
ncroxon, line 1: I'm good with this change.

super1.c
ncroxon, line 1: I'm good with this change.

udev.c
ncroxon, line 1: I'm good with this change.

util.c
ncroxon, line 1: I'm good with this change.

ncroxon (approved):
I have tested creating different array types, mkfs and mounting. Everything sync'd fine and was accessible.

mdadm.h
```diff
@@ -189,40 +197,6 @@ struct dlm_lksb {
...
-#define __cpu_to_be64(_x) (unsigned long long)(_x)
-#define	__be16_to_cpu(_x) (unsigned int)(_x)
-#define __be32_to_cpu(_x) (unsigned int)(_x)
-#define __be64_to_cpu(_x) (unsigned long long)(_x)
-#else
-#  error "unknown endianness."
-#endif
-#endif /* __KLIBC__ */
-
 /*
```
XiaoNi87, line 200: Could you explain about this part? We don't need to check klibc anymore?

super1.c
```diff
@@ -26,92 +26,6 @@
...
-
-	/* device state information. Indexed by dev_number.
-	 * 2 bytes per device
-	 * Note there are no per-device state flags. State information is rolled
-	 * into the 'roles' value.  If a device is spare or faulty, then it doesn't
-	 * have a meaningful role.
-	 */
-	__u16	dev_roles[0];	/* role in array, or 0xffff for a spare, or 0xfffe for faulty */
-};
-
```
XiaoNi87, line 114: The kernel header defined this too. This part looks good to me.

super1.c
```diff
@@ -147,19 +61,6 @@ struct misc_dev_info {
...
-					|MD_FEATURE_BAD_BLOCKS		\
-					|MD_FEATURE_REPLACEMENT		\
-					|MD_FEATURE_RESHAPE_BACKWARDS	\
-					|MD_FEATURE_NEW_OFFSET		\
-					|MD_FEATURE_BITMAP_VERSIONED	\
-					|MD_FEATURE_JOURNAL		\
-					|MD_FEATURE_PPL			\
-					|MD_FEATURE_MULTIPLE_PPLS	\
-					|MD_FEATURE_RAID0_LAYOUT	\
-					)
```
XiaoNi87, line 162: Maybe we can remove all feature_map bits here? Because kernel header defines these feature map bits too.

mdadm.h
```diff
@@ -162,8 +162,16 @@ struct dlm_lksb {
 #define GROW_SERVICE "mdadm-grow-continue"
 #endif /* GROW_SERVICE */
 
-#include	"md_u.h"
-#include	"md_p.h"
+#include	<linux/raid/md_u.h>
+#include	<linux/raid/md_p.h>
```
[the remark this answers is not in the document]
mtkaczyk: I just used IDE to find them, I did not expected issues. I think that it should be same even if placement is different, the most important it that `ld` knows how to find them.
In my case they are in `/usr/include/`

mdadm.h
mtkaczyk: I removed customized `le_to_cpu` functions therefore I don't think it is needed.

super1.c
mtkaczyk: Are you sure they are exported? I removed all defines that were conflicting with the kernel I have.
If they are not exported, we need to export them in kernel.
Maybe I should handle it this way then?
```c
#ifndef MD_FEATURE_RAID0_LAYOUT
#define MD_FEATURE_RAID0_LAYOUT		4096
#endif
```

For a while I cannot promise that I will do the patch to export them in kernel because I still don't have kernel development setup. If you have one and can export them I will be thankful!

Create.c
```diff
@@ -23,8 +23,8 @@
  */
 
 #include	"mdadm.h"
-#include	"md_u.h"
-#include	"md_p.h"
+#include	<linux/raid/md_u.h>
+#include	<linux/raid/md_p.h>
```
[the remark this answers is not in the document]
mtkaczyk: good one, we don't need. Thanks!.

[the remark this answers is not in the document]
mtkaczyk: I will respin it after we will clarify all questions.

super1.c
mtkaczyk: Hmm, looks like they are exported and I for some reasons was able to compile it.
Anyway, I removed them and renamed.

mtkaczyk:
@XiaoNi87 updated. I removed custom libc targets, in that case right way it to set CC. Not having klibc directly supported allowed me to removed implementation of custom `xx_to_cpu` functions and include `asm/byteorder.h`

mdadm.h
```diff
@@ -43,6 +43,7 @@ extern __off64_t lseek64 __P ((int __fd, __off64_t __offset, int __whence));
 #include	<sys/time.h>
 #include	<getopt.h>
 #include	<fcntl.h>
+#include	<ftw.h>
```
XiaoNi87, line 47: Hi Mariusz

Why do we need to include this head file?

XiaoNi87: I c. Because the parameter of function add_dev needs it. And the following change removes it. The change looks good to me. Thanks for your patch :)

super1.c
```diff
@@ -2643,7 +2523,7 @@ add_internal_bitmap1(struct supertype *st,
 	bms->nodes = __cpu_to_le32(st->nodes);
 	if (st->nodes)
 		sb->feature_map = __cpu_to_le32(__le32_to_cpu(sb->feature_map) |
-						MD_FEATURE_BITMAP_VERSIONED);
+						MD_FEATURE_CLUSTERED);
```
XiaoNi87, line 2526: It's better to describe this in commit message

mdadm.h
mtkaczyk: clarified offline

mtkaczyk:
Hey @neilbrown ,
Would you like to add something on this? Is there something we should know before we would merge this?

neilbrown:
I think this is sensible.  As stated, the current design made sense when we started, but makes a lot less sense now that the project is mature.
I doubt that klibc support was ever used much and if someone really wants to use a different libc, adding fresh support would probably be cleaner than sticking with what we have.

mtkaczyk:
@bkucman, @XiaoNi87 could yo please review this?

@neilbrown would you like to be added to the maintainers team on github?
Time to time you would give us your approval :)

neilbrown:
> @neilbrown would you like to be added to the maintainers team on github? Time to time you would give us your approval :)

No thanks.  I'm happy to provide advice on occasion (when asked) but I want to stay away from approval.

PR #154, opened 2025-03-11
super1: Clear extra flags when initializing metadata

wuguanghao3:
When adding a disk to a RAID1 array, the metadata is read from the existing member disks for sync. However, only the bad_blocks flag are copied, the bad_blocks records are not copied, so the bad_blocks records are all zeros. The kernel function super_1_load() detects bad_blocks flag and reads the bad_blocks record, then sets the bad block using badblocks_set().

After the kernel commit 1726c7746783 (badblocks: improve badblocks_set() for multiple ranges handling) if the length of a bad_blocks record is 0, it will return a failure. Therefore the device addition will fail.

So when adding a new disk, some flags cannot be sync and need to be clead.

### DIFF
--- a/super1.c
+++ b/super1.c
@@ -1971,6 +1971,9 @@ static int write_init_super1(struct supertype *st)
 	long bm_offset;
 	bool raid0_need_layout = false;
 
+	/* Clear extra flags */
+	sb->feature_map &= ~__cpu_to_le32(MD_FEATURE_BAD_BLOCKS | MD_FEATURE_REPLACEMENT);
+
 	/* Since linux kernel v5.4, raid0 always has a layout */
 	if (has_raid0_layout(sb) && get_linux_version() >= 5004000)
 		raid0_need_layout = true;
--- a/tests/05r1-add-badblocks
+++ b/tests/05r1-add-badblocks
@@ -1,0 +1,24 @@
+#
+# create a raid1 with a drive and set badblocks for the drive.
+# add a new drive does not cause an error.
+#
+
+# create raid1
+mdadm -CR $md0 -l1 -n2 -e1.0 $dev1 missing
+testdev $md0 1 $mdsize1a 64
+sleep 3
+
+# set badblocks for the drive
+dev1_name=$(basename $dev1)
+echo "100 100" > /sys/block/md0/md/dev-$dev1_name/bad_blocks
+echo "write_error" > /sys/block/md0/md/dev-$dev1_name/state
+
+# write badblocks to metadata
+dd if=/dev/zero of=$md0 bs=512 count=200 oflag=direct
+
+# re-add and recovery
+mdadm $md0 -a $dev2
+check recovery
+
+mdadm -S $md0
+


XiaoNi87:
For upstream tests, I'll fix it ASAP. For the checkpatch errors, please fix them

wuguanghao3:
> For upstream tests, I'll fix it ASAP. For the checkpatch errors, please fix them

This error is likely a false positive. The commit id is for a kernel commit, not mdadm, so it cannot be detected.

XiaoNi87:
Thanks for this explanation. So this patch looks good to me. Upstream tests failure should be fixed today.

super1.c
```diff
@@ -1971,6 +1971,9 @@ static int write_init_super1(struct supertype *st)
 	long bm_offset;
 	bool raid0_need_layout = false;
 
+	/* Clear extra flags */
+	sb->feature_map &= ~__cpu_to_le32(MD_FEATURE_BAD_BLOCKS | MD_FEATURE_REPLACEMENT);
```
mtkaczyk, line 1975: Question here, how this will affect kernels without mentioned commit?

tests/05r1-add-badblocks
```diff
@@ -0,0 +1,25 @@
...
+sleep 3
+
+# set badblocks for the drive
+dev1_name=$(basename $dev1)
+echo "10000 100" > /sys/block/md0/md/dev-$dev1_name/bad_blocks
+echo "write_error" > /sys/block/md0/md/dev-$dev1_name/state
+
+# maybe fail but that's ok, as it's intended to 
+# record the bad block in the metadata.
+mkfs.ext3 $md0
```
mtkaczyk, line 18: I don't like using filesystem logic to record badblock. can we do it easier, preferably by `dd`?
Is it based on other test?

mtkaczyk (commented):
Hi @wuguanghao3 
thanks for detailed explanation. I have two questions to this. 
First one is about backward compatibility, I think it is important to verify it and add to commit description.

tests/05r1-add-badblocks
wuguanghao3: > I don't like using filesystem logic to record badblock. can we do it easier, preferably by `dd`? Is it based on other test?

`dd` should also work. The purpose of `mkfs` or other operations is just to write the manually set bad blocks to the disk.

super1.c
wuguanghao3: > Question here, how this will affect kernels without mentioned commit?

There will be no impact. The kernel has not merged the commit for length validation, while mdadm has merged this commit, which actually reduces the setting of invalid bad_blocks. Because if this commit is not merged, the kernel will set 512 bad block records [start:0, len:0].

tests/05r1-add-badblocks
mtkaczyk: please use dd then instead. Simply, with dd we are immediately see that rages of bbs and area we are writing matches. Filesystem creation has own logic we don't probably need to care about at all.

wuguanghao3: OK, I will replace it with the `dd`

XiaoNi87:
This look good to me

XiaoNi87:
Hi Guanghao

Maybe you need to re-submit your patches based on the latest version of https://github.com/md-raid-utilities/mdadm. So your PR can pass regression tests. I submitted a patch set which can make regression pass.

bkucman:
> Hi Guanghao
> 
> Maybe you need to re-submit your patches based on the latest version of https://github.com/md-raid-utilities/mdadm. So your PR can pass regression tests. I submitted a patch set which can make regression pass.

Hi,
Yes, to run fixed tests here, you need to rebase the change to upstream.

wuguanghao3:
> Hi Guanghao
> 
> Maybe you need to re-submit your patches based on the latest version of https://github.com/md-raid-utilities/mdadm. So your PR can pass regression tests. I submitted a patch set which can make regression pass.

OK. I will re-submit later.

XiaoNi87 (approved):
Now only 10ddf-fail-two-spares fails. I can't reproduce this in my machine. But this PR doesn't change anything of ddf. I'm good to this change.

mtkaczyk (approved):

PR #156, opened 2025-03-12
Regression fix

XiaoNi87:


tests/skiptests
```diff
@@ -0,0 +1,9 @@
+casename:seconds
+01raid6integ:1732
+01replace:396
+07layouts:836
+11spare-migration:1140
+12imsm-r0_2d-grow-r0_5d:218
+13imsm-r0_r0_2d-grow-r0_r0_4d:218
+13imsm-r0_r0_2d-grow-r0_r0_5d:246
+19raid6check:268
```
mtkaczyk, line 9: This is nice but did you considered using some tag/label mechanism which is popular solution to handle such cases?
For example, we can consider adding labels to test and then run some tests with label, or without label, or mixed.
(2 labels defined, one label defined one not defined etc.).

Labels could be "imsm", "native", "long_test", "md_regression" or even "broken"

Just thinking, it could be an general answer for test filtering at all.

mtkaczyk (commented):
Hi Xiao,
One question, see in review.

LGTM.

tests/skiptests
XiaoNi87: Hi Mariusz

It's a good suggestion. We can take it as a future feature.  Now I want to resolve the ci testing failure ASAP. So the PRs can't be stuck because of ci testing failure. 

But as you see, the ci test fails again :(

ncroxon (approved):

ncroxon (approved):

ncroxon (approved):

PR #159, opened 2025-03-19
mdadm: check posix name before setting name and devname

XiaoNi87:
It's good to has limitations for name when creating an array. But the arrays which were created before patch e2eb503bd797 (mdadm: Follow POSIX Portable Character Set) can't be assembled. So remove the POSIX check for assemble mode.

This can be reproduced:
* build mdadm without patch e2eb503bd797
* mdadm -CR /dev/md/node1:pv1 -l0 -n2 /dev/loop0 /dev/loop1
* mdadm -Ss
* build with latest mdadm, and try to assemble it.
* mdadm -A /dev/md/node1:pv1 --name node1:pv1

Fixes: e2eb503bd797 (mdadm: Follow POSIX Portable Character Set)

### DIFF
--- a/config.c
+++ b/config.c
@@ -208,11 +208,6 @@ static mdadm_status_t ident_check_name(const char *name, const char *prop_name, 
 		return MDADM_STATUS_ERROR;
 	}
 
-	if (!is_name_posix_compatible(name)) {
-		ident_log(prop_name, name, "Not POSIX compatible", cmdline);
-		return MDADM_STATUS_ERROR;
-	}
-
 	return MDADM_STATUS_SUCCESS;
 }
 
@@ -512,7 +507,8 @@ void arrayline(char *line)
 
 	for (w = dl_next(line); w != line; w = dl_next(w)) {
 		if (w[0] == '/' || strchr(w, '=') == NULL) {
-			_ident_set_devname(&mis, w, false);
+			if (is_name_posix_compatible(basename(w)))
+				_ident_set_devname(&mis, w, false);
 		} else if (strncasecmp(w, "uuid=", 5) == 0) {
 			if (mis.uuid_set)
 				pr_err("only specify uuid once, %s ignored.\n",
--- a/mdadm.c
+++ b/mdadm.c
@@ -732,6 +732,12 @@ int main(int argc, char *argv[])
 				exit(2);
 			}
 
+			if (mode != ASSEMBLE &&
+					!is_name_posix_compatible(basename(optarg))) {
+				pr_err("%s Not POSIX compatible\n", basename(optarg));
+				exit(2);
+			}
+
 			if (ident_set_name(&ident, optarg) != MDADM_STATUS_SUCCESS)
 				exit(2);
 
@@ -1284,11 +1290,18 @@ int main(int argc, char *argv[])
 	    mode == GROW || (mode == ASSEMBLE && ! c.scan)) {
 		struct stat stb;
 		int ret;
+		char *bname = basename(devlist->devname);
 
 		if (devs_found < 1) {
 			pr_err("an md device must be given in this mode\n");
 			exit(2);
 		}
+
+		if (mode != ASSEMBLE && !is_name_posix_compatible(bname)) {
+			pr_err("%s Not POSIX compatible\n", bname);
+			exit(2);
+		}
+
 		if (ident_set_devname(&ident, devlist->devname) != MDADM_STATUS_SUCCESS)
 			exit(1);
 


XiaoNi87:
* rm -rf /var/run/mdadm/map
* mdadm -A /dev/md/node1:pv1 --name node11:pv43
* cat /proc/mdstat
Personalities : [raid6] [raid5] [raid4] [raid0] [raid1] [raid10]
unused devices: <none>
* mdadm -A /dev/md/node1:pv1 --name node1:pv1
mdadm: /dev/md/node1:pv1 has been started with 2 drives.

So --name is used to find the right disk to assemble the array

mtkaczyk:
> * rm -rf /var/run/mdadm/map
> 
>     * mdadm -A /dev/md/node1:pv1 --name node11:pv43
> 
>     * cat /proc/mdstat
>       Personalities : [raid6] [raid5] [raid4] [raid0] [raid1] [raid10]
>       unused devices:

We should look for the array with name `node1:pv1` in metadata and assemble it as `node11:pv43`. That is what would have some sense for me.

Please also try:
```
#mdadm -A /dev/md/node11:pv13 --name node1:pv1
```

It this is also not working then IMO it is weirdo that should gone. What is the point to pass same name 2 times?

I understand that someone may use this and it is a functional regression so here  I asking what is the use case? Can you share?

XiaoNi87:
Hi Mariusz

[root@ mdadm]# mdadm -A /dev/md/node11:pv13 --name node1:pv1
mdadm: /dev/md/node11:pv13 has been started with 2 drives.
[root@ mdadm]# ls /dev/md/node11\:pv13
/dev/md/node11:pv13

This problem is reported by one customer. It's the reason why I tried to fix this. But I don't know the details. I'll ask.

For the two names, it should make sense. One is device node name and the other is metadata name, right? And as you see, if it doesn't specify --name, it can't assemble the array. Because the customer doesn't use conf file.

Thanks
Xiao

AdamWill:
I think the real problem here is that `is_name_posix_compatible` has no business blocking the `:` character, since *mdadm itself* frequently introduces that character into device names (see `update_super1` and `init_super1` blocks where the system's hostname is prepended to the device name and a `:` separator added).

This check should allow :. And it shouldn't be called `is_name_posix_compatible`, because it's not checking that at all, it disallows many characters that are part of the POSIX character set.

I've sent https://github.com/md-raid-utilities/mdadm/pull/160 to rename the function and have it allow `:`.

AdamWill:
...although I do kinda agree with @XiaoNi87 that this check is also problematic for any case where someone happened to use an older mdadm to create a device with a name containing characters which this check blocks. If the problem we had was 'device names with a space character in them', why doesn't this check *only* refuse whitespace? Why get into the business of blocking other characters? I can imagine someone might have put a comma in a device name, for instance.

mtkaczyk:
Thanks, now I see a problem. Our --name is device determiner of array to be assembled.

We need to allow anything there and avoid POSIX check for name passed in assemble. That is for sure. 

However I'm in doubt with devname (the "/dev/md/node1:name1") parameter. It should be forbidden.
User should not specify homehost directly I think. @XiaoNi87 what do you think?

neilbrown:
> However I'm in doubt with devname (the "/dev/md/node1:name1") parameter. It should be forbidden. User should not specify homehost directly I think. @XiaoNi87 what do you think?

I think that validating the device name like this is a bad idea.  Generating a warning might be useful if the name is known to cause problems for udev, but maybe the user doesn't care about udev.
Unix/linux generally lets you do whatever you want and only stops you from doing things that are impossible or dangerous.

It might be valid to prevent spaces tabs and newlines in the "name" recorded on metadata, because mdadm uses these as separators when parsing the config file.  Anything more seems excessive.  "Posix compatible" is not a valid reason.  "Posix" is for portable applications.  mdadm is Linux-only. It doesn't need to be portable anywhere else.  Compatibility with udev is a valid concern but problems should only result in a warning which explicitly mentions udev.

I note that none of these rules about valid names are mentioned in the man pages.

Banning the leading dot is also weird.

"Having leading dot might be confusing, MD device cannot be hidden."

That is not a good reason for stopping and admin from doing what they choose to do.  mdadm is a tool for people who know what they are doing.  When writing a GUI to manage md/raid it might be perfectly appropriate to put limits like this on names.  It is not at all appropriate to do it in mdadm itself.

Please consider reverting these name checks.  They are obviously causing problems and there is not clear statement of benefit.

mtkaczyk:
Thanks @neilbrown  for the comment.

> > However I'm in doubt with devname (the "/dev/md/node1:name1") parameter. It should be forbidden. User should not specify homehost directly I think. @XiaoNi87 what do you think?
> 
> I think that validating the device name like this is a bad idea. Generating a warning might be useful if the name is known to cause problems for udev, but maybe the user doesn't care about udev. Unix/linux generally lets you do whatever you want and only stops you from doing things that are impossible or dangerous.

Hmm. I never thought about that this way. Basically you are saying we can release requirements and tolerate unexpected scenarios. I wanted to close them as much as possible and allow only what is consistent. mdadm is old tool and having  discussion about POSIX I implemented make me understanding that your approach is safer.

I must also admit that now I'm not sponsored by Intel anymore to work on mdadm and I prefer to allow more rather than resolving issues caused by my changes.

> It might be valid to prevent spaces tabs and newlines in the "name" recorded on metadata, because mdadm uses these as separators when parsing the config file. Anything more seems excessive. "Posix compatible" is not a valid reason. "Posix" is for portable applications. mdadm is Linux-only. It doesn't need to be portable anywhere else. Compatibility with udev is a valid concern but problems should only result in a warning which explicitly mentions udev.
> 
Please kindly note that we support IMSM and IMSM is portable. My primary motivation was to close differences in various IMSM drivers (UEFI, WINDOWS) to keep compatibility. Initially, I submitted it in this shape to get feedback as I was not sure how to implemented this and it got merged.

But.. I don't see it as a requirement anymore as IMSM is in sustaining.

> I note that none of these rules about valid names are mentioned in the man pages.
> 
> Banning the leading dot is also weird.
> 
> "Having leading dot might be confusing, MD device cannot be hidden."

> 
> That is not a good reason for stopping and admin from doing what they choose to do. mdadm is a tool for people who know what they are doing. When writing a GUI to manage md/raid it might be perfectly appropriate to put limits like this on names. It is not at all appropriate to do it in mdadm itself.

Yes, it was reported by validation team that validated it like GUI. I never thought about that this way for sure.
My approach was - if something has no sense, block it to not keep the reason in mind.

> 
> Please consider reverting these name checks. They are obviously causing problems and there is not clear statement of benefit.

Overall I'm fine with that. I don't have motivation to keep it from various reasons, especially limited time to deal with issues. You are saing that we should keep it in Linux style base tool, even if in the past I wanted to achive more then it gone now. Let me make some patches.

Thanks Neil, you helped a lot to clear my brain. I was unsure what approach take because I did this more than 2 years ago from now and I didn't saw good reason why I need to start reverting. 

Thinking about this like about base utility that gives unlimited power is helpful here.

XiaoNi87:
@neilbrown Hi Neil, thanks very much for your help. I close this one because @mtkaczyk decided to revert the change which has ASCII check https://github.com/md-raid-utilities/mdadm/pull/165

PR #160, opened 2025-03-21
Rename is_name_posix_compatible to is_name_valid, allow :

AdamWill:
The name of this function is a lie. The POSIX Portable Character Set contains far more characters than just letters, digits and _ . and -. Notably, it contains the space character, which this function was specifically introduced to disallow.

The function also disallows the : character, which is a problem because we often add this character to device names ourselves, when we prepend the hostname. So with this check it is possible to be blocked from assembling a device under the name mdadm gave it:

mdadm --assemble /dev/md/system2:pv00 --run --uuid=8e4473bc:997208c1:da327078:0dc4d134 /dev/sda3 /dev/sdb1 mdadm: Value "system2:pv00" cannot be set as devname. Reason: Not POSIX compatible.

So, let's give this function an accurate name, and relax it to allow the : character.

### DIFF
--- a/config.c
+++ b/config.c
@@ -193,7 +193,7 @@ inline void ident_init(struct mddev_ident *ident)
  * @prop_name: the name of the property it is validated against, used for logging.
  * @cmdline: context dependent actions.
  *
- * @name must follow name's criteria, be POSIX compatible and does not have leading dot.
+ * @name must follow name's criteria, contain no forbidden characters and not have leading dot.
  */
 static mdadm_status_t ident_check_name(const char *name, const char *prop_name, const bool cmdline)
 {
@@ -208,8 +208,8 @@ static mdadm_status_t ident_check_name(const char *name, const char *prop_name, 
 		return MDADM_STATUS_ERROR;
 	}
 
-	if (!is_name_posix_compatible(name)) {
-		ident_log(prop_name, name, "Not POSIX compatible", cmdline);
+	if (!is_name_valid(name)) {
+		ident_log(prop_name, name, "Contains forbidden characters", cmdline);
 		return MDADM_STATUS_ERROR;
 	}
 
--- a/lib.c
+++ b/lib.c
@@ -472,20 +472,20 @@ bool is_alphanum(const char c)
 }
 
 /**
- * is_name_posix_compatible() - Check if name is POSIX compatible.
+ * is_name_valid() - Check if name contains disallowed characters.
  * @name: name to check.
  *
- *  POSIX portable file name character set contains ASCII letters,
- *  digits, '_', '.', and '-'. Also forbid leading '-'.
+ *  Allows name to contain only ASCII letters, digits, '_', '.', ':' and '-'.
+ *  Also forbid leading '-'.
  *  The length of the name cannot exceed NAME_MAX - 1 (ensure NULL ending).
  *
  * Return: %true on success, %false otherwise.
  */
-bool is_name_posix_compatible(const char * const name)
+bool is_name_valid(const char * const name)
 {
 	assert(name);
 
-	char allowed_symbols[] = "-_.";
+	char allowed_symbols[] = "-_.:";
 	const char *n = name;
 
 	if (!is_string_lq(name, NAME_MAX))
--- a/mdadm.8.in
+++ b/mdadm.8.in
@@ -885,7 +885,7 @@ are used to add different devices).
 Set a
 .B name
 for the array. It must be
-.BR "POSIX PORTABLE NAME"
+.BR "MDADM NAME FORMAT"
 compatible and cannot be longer than 32 chars. This is effective when creating an array
 with a v1 metadata, or an external array.
 
@@ -1025,7 +1025,7 @@ is much safer.
 .TP
 .BR \-N ", " \-\-name=
 Specify the name of the array to assemble. It must be
-.BR "POSIX PORTABLE NAME"
+.BR "MDADM NAME FORMAT"
 compatible and cannot be longer than 32 chars. This must be the name
 that was specified when creating the array. It must either match
 the name stored in the superblock exactly, or it must match
@@ -2237,7 +2237,7 @@ and
 The
 .B name
 option updates the subarray name in the metadata. It must be
-.BR "POSIX PORTABLE NAME"
+.BR "MDADM NAME FORMAT"
 compatible and cannot be longer than 32 chars. If successes, new value will be respected after
 next assembly.
 
@@ -3189,8 +3189,8 @@ When
 .B \-\-incremental
 mode is used, this file gets a list of arrays currently being created.
 
-.SH POSIX PORTABLE NAME
-A valid name can only consist of characters "A-Za-z0-9.-_".
+.SH MDADM NAME FORMAT
+A valid name can only consist of characters "A-Za-z0-9.-_:".
 The name cannot start with a leading "-" and cannot exceed 255 chars.
 
 .SH DEVICE NAMES
@@ -3215,7 +3215,7 @@ can be given, or just the suffix of the second sort of name, such as
 can be given.
 
 In every style, raw name must be compatible with
-.BR "POSIX PORTABLE NAME"
+.BR "MDADM NAME FORMAT"
 and has to be no longer than 32 chars.
 
 When
--- a/mdadm.h
+++ b/mdadm.h
@@ -1675,7 +1675,7 @@ extern int check_raid(int fd, char *name);
 extern int check_partitions(int fd, char *dname,
 			    unsigned long long freesize,
 			    unsigned long long size);
-extern bool is_name_posix_compatible(const char *path);
+extern bool is_name_valid(const char *path);
 extern int fstat_is_blkdev(int fd, char *devname, dev_t *rdev);
 extern int stat_is_blkdev(char *devname, dev_t *rdev);
 


AdamWill:
Note, I'm not sure if renaming it is an API change (I'm not a trained C coder and don't entirely understand hwo to figure out if a C function is part of public API or not). If so, we could keep an inline `is_name_posix_compatible` which just calls `is_name_valid` to preserve compatibility, I guess. Let me know if you want me to add that.

AdamWill:
Note the blocks in super1.c `update_super1` and `init_super1` which add the `:` character to device names: https://github.com/md-raid-utilities/mdadm/blob/ee3a6cab09c8acaf6706b3710f5652e9be43b57e/super1.c#L1664-L1678 and https://github.com/md-raid-utilities/mdadm/blob/ee3a6cab09c8acaf6706b3710f5652e9be43b57e/super1.c#L1226-L1238 .

This erroneous refusal to assemble devices under the name mdadm gave them is causing a release-blocking bug in Fedora: https://bugzilla.redhat.com/show_bug.cgi?id=2325906

AdamWill:
checkpatch doesn't like the long lines in the commit message, but as they're the actual command and the error message it seems hard to shorten them, I don't really want to wrap them.

AdamWill:
POSIX Portable Character Set: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap06.html#tag_06_01 . It includes tab, newline, vertical tab, form feed, carriage return and space. And the colon. And the comma.

AdamWill:
Looking at udev docs, they say:

"The set of characters to name a symlink is limited. Allowed characters are [0-9A-Za-z#+-.:=@_/], valid utf8 character sequences, and "\x00" hex encoding. All other characters are replaced by a _ character."

so perhaps what we really ought to do is mirror that? allow those characters, and treat this not as a validity check but as a sanitization operation - the function accepts a string and sanitizes it according to those rules, then returns it?

AdamWill:
I note there was some discussion of this problem in https://lore.kernel.org/linux-raid/ff7cc9e9-532b-5a23-cdb7-e6b4521a2004@trained-monkey.org/T/#t and Mariusz claimed "For sure it will be fixed before release.", but...it doesn't seem like it was?

mtkaczyk:
Hi @AdamWill 
Thanks for detailed analysis. I will look into int more deeper in few days.

This is not simple topic. At some point, I took a challenge to be more strict and now users are paying  in regressions.
It is already there so before we will consider reverting I must to be double sure that there is no other way.

I have various reasons to remove ":" from the allowed set, but to be honest at that point I was more IMSM focused rather than native.

AdamWill:
well, the alternative would be to stop using `:` as a separator when producing names for devices ourselves, I guess. But creating devices with `:` in their names and then refusing to assemble them is clearly a bug.

XiaoNi87:
Hi Adam

My pull request only tries to resolve the issue https://issues.redhat.com/browse/RHEL-72756. At first, I want to do the same as you do. But Mariusz's explanation makes sense to me. We can use --homehost to give the hostname. 

For example:
mdadm -CR /dev/md/vol1 -l0 -n2 /dev/loop0 /dev/loop1 --homehost node1
[root@ mdadm]# mdadm -D /dev/md/vol1 | grep Name
              Name : node1:vol1
The metatdata name has the hostname

And to make things simple, I prefer disabling ":" in name and keep it as it is now. Now the code is too complex to analyze a name. It needs to consider many situations. 

But I agree with your point that the name of the function should change. 

Thanks
Xiao

XiaoNi87:
By the way, the incremental assemble work well.
mdadm -CR /dev/md/vol1 -l0 -n2 /dev/loop0 /dev/loop1 --homehost node1
mdadm -Ss
[root@ mdadm]# mdadm -I /dev/loop0
mdadm: /dev/loop0 attached to /dev/md/node1:vol1, not enough to start (1).
[root@ mdadm]# mdadm -I /dev/loop1
mdadm: /dev/loop1 attached to /dev/md/node1:vol1, which has been started.
[root@ mdadm]# ll /dev/md/node1\:vol1
lrwxrwxrwx. 1 root root 8 Mar 25 05:27 /dev/md/node1:vol1 -> ../md127

mtkaczyk:
> well, the alternative would be to stop using : as a separator when producing names for devices ourselves, I guess. But creating devices with : in their names and then refusing to assemble them is clearly a bug.

I cannot change how it was implemented. What I did, was blocking users from creating such arrays and because code is generic it affected multiple scenarios. Your case is affected because the usecase you have I determined as invalid. I will explain it in details below.

I decided to take a risk and potentially introduce a regression for someone with extraordinary case because without this approach I would never close this (there are users that are having arrays for more than 10 years).

> The mysteriousness here seems a bit weird. What was the reason for the "challenge" you took? What are the "various reasons" you have to block :? This would all be simpler if this was written down. The only clear explanation provided for the introduction of this check was an issue with udev not allowing whitespace in names.

Sure, you deserved some explanation.  Please note that it is a legacy product and some functionalities are there for years but we are continuously dealing with them. That is why not all might have sense for you. It is how it is.

We have support for 3 metadata formats. Each format has quite different implementation. In case of native raid `:` is used in metadata name for spiting the name into two parts "homehost:name"

homehost is the hostname of system that the array was initially created. The most trivial usecase is no need for mdadm.conf to determine locality and assemble array as local (we are not appending "_0, _1" in /dev/md/).

Why it matters? there are some points I considered:
- generic and metadata independent code for parsing the name
-  highlighting special meaning of ":" for native metadata
- keeping the interface consistent for every metadata

and here is my way of thinking:

-------------
Because ":" has special meaning for one metadata type, it must be possibly  the most visible for users. Therefore It should be reserved only for native raid arrays to prevent users from abusing it . It is not same in other metadata types, so banning ":" is fast way to prevent users from seeing functional differences for example in case if they are using many metadata formats.

It should not be also part of the name user is choosing to give user a clue it is setting something else, that why it is allowed with  --homehost now.

And finally, the code. I blocked it globally to minimize the size of the change. I implemented functionality that is consistent however, I also said in commit message that I'm aware of regression in:
https://web.git.kernel.org/pub/scm/utils/mdadm/mdadm.git/commit/?id=e2eb503bd797908f515b58428b274f1ba6a05349

--------------------------

> to be clear on the case we have: our installer needs to find and then potentially assemble RAID devices present on the disks. they, obviously, were not created within that OS, so there is no configuration related to them present. how it's doing this is to get info about the RAID devices from udev, and use the MD_DEVNAME property in the udev data to determine their location in /dev. This problem is breaking that approach. We get an MD_DEVNAME name like localhost-live:pv00, so we try to activate the device at /dev/md/localhost-live:pv00 - which is a perfectly correct location - and it fails because of this problem.

Thanks for pointing real life scenario that needs to be addressed. In Suse or Redhat it is solved this way:
```
# Scan all block devices, generate raid config and save it
mdadm -Dbs > /tmp/mdadm.conf or mdadm -Ebs > /tmp/mdadm.conf 

# Assemble arrays that are described in config
mdadm -As -c /tmp/mdadm.conf
```

Please check if it is able to handle your problem. Please also note that --assemble in mdadm is terrible. I would recommend implementing new functionality, basing on --Incremental because it is simply more reliable and better tested (udev is using incremental not assemble).

For you code, I will look into is after we will make a deal and agree on the approach :)

Anyway, I still think that  ":" is special and you should not use it when assembling.  I know that you are taking in from MD_DEVNAME but it is still a way you may set different name than the one in metadata.

I'm closre to say that there should be no need to specify devname when assembling because we can get that from metadata. Please kindly check if the WA I provided above satisfy you needs. If not, I will start thinking on something else.

I'm always open to change my mind to satisfy users. First, we need to have this discussion to understand if we really need to dig into legacy scenario that can be replaced easy.

AdamWill:
I just want to keep this really simple: the issue we have is that, entirely internally - this does not involve the user doing anything specific at all - mdadm is deciding to name devices with names that it then refuses to interact with because it deems them "invalid".

We, the external user of mdadm, are not putting the `:` in the name. mdadm is naming the device itself. If mdadm didn't put a `:` in the name, there wouldn't be a problem.

This name is exposed in the metadata you can access in various ways (including udev) and also *as an actual filename*. We get a `/dev/md/something:somethingelse`. All I'm saying is that mdadm should *either*:

1) allow operations on the device under the name *mdadm itself gave it* (i.e. change the check to allow `:`, as in this PR, or somehow not apply this check to any operation you might reasonably do based on this name), or
2) create devices with a different name that is accepted by this check (i.e. use an allowed character as the separator when deciding what to name devices).

mtkaczyk:
> We, the external user of mdadm, are not putting the : in the name. mdadm is naming the device itself. If mdadm didn't put a : in the name, there wouldn't be a problem.

You can achieve this by making the generated string "hostname:name" longer than 32 :)
So you would change your hostname or simply made a name longer.  Mind a try?

As I understand, you are using command like this:
mdadm -A /dev/md/node1:pv1 --name node1:pv1

And you would like to not touch this functionality in your installer, correct?

AdamWill:
I mean, we can change how the installer does things, though we would prefer not to have to do that for Fedora 42 since it's already in stabilization for Final. But it seems like a weird thing to ask, because these are all *longstanding features of mdadm*. We aren't doing anything crazy and off-the-books, here. They seem like things people might perfectly well be likely to do in other contexts, too, and want it to work.

I just don't understand the complicated response to this issue, which seems like a really, really simple issue. mdadm is all one codebase. It has these features which are clearly in conflict with each other. To me that is clearly a bug in mdadm. I would find it weird to be maintaining a tool which had one bit which gave things names in a given format, and another bit which refused to interact with them using those names. Don't you? Isn't it weird to be suggesting workarounds by which we can game mdadm's naming logic to leave out the problematic character, instead of just fixing mdadm so it doesn't consider the character problematic, *or* doesn't include it in names ever?

AdamWill:
Oh, sorry, missed something from an earlier comment:

> In Suse or Redhat it is solved this way

To be clear, I'm the Fedora QA team lead. I work for Red Hat. The installer I'm talking about is the Red Hat / Fedora installer. That's the installer that's broken by this change.

AdamWill:
so...I'm re-reading your comments here and trying to understand the objection. As best as I can tell, what you want to block is people including `:` when *manually naming devices*, right? Which is not what we're doing here, AFAICS. The command we run looks like:
```
mdadm --assemble /dev/md/system2:pv00 --run --uuid=8e4473bc:997208c1:da327078:0dc4d134 /dev/sda3 /dev/sdb1
```
we're not *naming* the device `system2:pv00`, here. It's *already* named that. `/dev/md/system2:pv00 ` already exists, and we did not give it that name.

But I guess the problem is that, for reasons I don't fully understand, it looks like this codepath goes through a function which seems like it's for *setting* a device name - `ident_set_name`. It happens [here in mdadm.c](https://github.com/md-raid-utilities/mdadm/blob/4e2e208c8d3e9ba0fae88136d7c4cd0292af73b0/mdadm.c#L727-L738).

So...is there an alternative option here where we somehow make that not happen? If this check only applied when you were really *setting* names, as opposed to on paths where you're referring to an existing device by a name you didn't set yourself, it'd be much less of an issue.

AdamWill:
I can't really write a PR for this myself as I just don't understand the codebase well enough, but what I'm envisioning is maybe if `ident_set_devname` and/or `ident_set_name` had a kinda preflight check where they see if the device in question *already exists under the name specified*, and if so, they skip `ident_check_name`?

XiaoNi87:
> 
> We, the external user of mdadm, are not putting the `:` in the name. mdadm is naming the device itself. If mdadm didn't put a `:` in the name, there wouldn't be a problem.

Hi Adam

Do you know what's the create command? I don't know why mdadm can create a name (/dev/md/system2:pv00) automatically. I tried myself, but I couldn't see this.

Thanks
Xiao

XiaoNi87:
> I just want to keep this really simple: the issue we have is that, entirely internally - this does not involve the user doing anything specific at all - mdadm is deciding to name devices with names that it then refuses to interact with because it deems them "invalid".
> 
> We, the external user of mdadm, are not putting the `:` in the name. mdadm is naming the device itself. If mdadm didn't put a `:` in the name, there wouldn't be a problem.

Hi Adam

Do you know what's the create command? I don't know why mdadm can create a name (/dev/md/system2:pv00) automatically. I tried myself, but I couldn't see this. If the create command doesn't specify `:` in device name, it should not create such name.

XiaoNi87:
> so...I'm re-reading your comments here and trying to understand the objection. As best as I can tell, what you want to block is people including `:` when _manually naming devices_, right? Which is not what we're doing here, AFAICS. The command we run looks like:
> 
> ```
> mdadm --assemble /dev/md/system2:pv00 --run --uuid=8e4473bc:997208c1:da327078:0dc4d134 /dev/sda3 /dev/sdb1
> ```
> 
> we're not _naming_ the device `system2:pv00`, here. It's _already_ named that. `/dev/md/system2:pv00` already exists, and we did not give it that name. (Admittedly, I'm now asking myself what 'assembling' the device actually _means_ in this context; what does this command _do_, if `/dev/md/system2:pv00` already exists?)

Ok, so we're talking the same problem which I reported at https://issues.redhat.com/browse/RHEL-72756. And my original thought is only to fix the assemble failure problem. But there is a unfortunate thing, the assemble command can change the device node name, for example:

mdadm -CR /dev/md0 -l0 -n2 /dev/loop0 /dev/loop1
mdadm -Ss
[root@]# mdadm -A /dev/md/node1:lv1 /dev/loop0 /dev/loop1
mdadm: /dev/md/node1:lv1 has been started with 2 drives.

Now the array is assembled with a new name which has ':'. My PR is used to remove the limitation for assemble mode, but it also can give the opportunity that users can create a new device name with ":". Sure I think most users don't do such things. 

> 
> But I guess the problem is that, for reasons I don't fully understand, it looks like this codepath goes through a function which seems like it's for _setting_ a device name - `ident_set_name`. It happens [here in mdadm.c](https://github.com/md-raid-utilities/mdadm/blob/4e2e208c8d3e9ba0fae88136d7c4cd0292af73b0/mdadm.c#L727-L738). **EDIT**: or maybe we're actually [here, uisng `ident_set_devname`](https://github.com/md-raid-utilities/mdadm/blob/4e2e208c8d3e9ba0fae88136d7c4cd0292af73b0/mdadm.c#L1292)? Either way it's kinda the same, I think, we wind up calling `ident_check_name`...
> 
> So...is there an alternative option here where we somehow make that not happen? If this check only applied when you were really _setting_ names, as opposed to on paths where you're referring to an existing device by a name you didn't set yourself, it'd be much less of an issue.
> 
> edit: hmm, the manpage notes that `mdadm --assemble` will create the dev node if udev isn't active, so is this related to that?

https://github.com/md-raid-utilities/mdadm/pull/159 please check here

Thanks
Xiao

XiaoNi87:
> > Do you know what's the create command? I don't know why mdadm can create a name (/dev/md/system2:pv00) automatically.
> 
> See [here](https://github.com/md-raid-utilities/mdadm/blob/4e2e208c8d3e9ba0fae88136d7c4cd0292af73b0/super1.c#L1664-L1669) and [here](https://github.com/md-raid-utilities/mdadm/blob/4e2e208c8d3e9ba0fae88136d7c4cd0292af73b0/super1.c#L1226-L1231) in super1.c.

Hi Adam

This is metadata name rather than device node name. For example:
env hostname: node1
mdadm -CR vol0 -l0 -n2 /dev/loop0 /dev/loop1
(mdadm: array /dev/md/vol0 started.)
mdadm -CR vol0 -l0 -n2 /dev/loop0 /dev/loop1 --name node1
(mdadm: array /dev/md/vol0 started.)

I guess libvet should specify the device name system2:pv00 itself when creating an array. In fedora40/41, mdadm allows to create an array which device node name has ':'. But in fedora42 installation, it doesn't allow to assemble the array which has been used in fedora40/41. 

> 
> > #159 please check here
> 
> Yeah, I saw that, but it's not quite the same. As you say, that just removes the check for assemble mode entirely, but we've both mentioned cases where assemble mode _does_ create the device, and we might want to keep the check in those cases. That's why I suggested instead having logic in `_ident_set_devname` to skip the `ident_check_name` check _if the device already exists_. but it's only a suggestion; you might be able to see a problem with this.

Hmm, I'm confused here. We always think an array already exists when talking about assemble mode. Why do you see assemble mode can create an array?

XiaoNi87:
> > Do you know what's the create command? I don't know why mdadm can create a name (/dev/md/system2:pv00) automatically.
> 
> See [here](https://github.com/md-raid-utilities/mdadm/blob/4e2e208c8d3e9ba0fae88136d7c4cd0292af73b0/super1.c#L1664-L1669) and [here](https://github.com/md-raid-utilities/mdadm/blob/4e2e208c8d3e9ba0fae88136d7c4cd0292af73b0/super1.c#L1226-L1231) in super1.c.

Hi Adam

This is metadata name rather than device node name. For example:
env hostname: node1
mdadm -CR vol0 -l0 -n2 /dev/loop0 /dev/loop1
(mdadm: array /dev/md/vol0 started.)
mdadm -CR vol0 -l0 -n2 /dev/loop0 /dev/loop1 --name node1
(mdadm: array /dev/md/vol0 started.)

[root@mdadm]# mdadm -D --export /dev/md/vol0
MD_DEVNAME=vol0 (Now we're talking the problem of MD_DEVNAME)
MD_NAME=node1:node1

I guess libvet should specify the device name system2:pv00 itself when creating an array. In fedora40/41, mdadm allows to create an array which device node name has ':'. But in fedora42 installation, it doesn't allow to assemble the array which has been used in fedora40/41.

> 
> > #159 please check here
> 
> Yeah, I saw that, but it's not quite the same. As you say, that just removes the check for assemble mode entirely, but we've both mentioned cases where assemble mode _does_ create the device, and we might want to keep the check in those cases. That's why I suggested instead having logic in `_ident_set_devname` to skip the `ident_check_name` check _if the device already exists_. but it's only a suggestion; you might be able to see a problem with this.

Hmm, I'm confused here. We always think an array already exists when talking about assemble mode. Why do you see assemble mode can create an array?

AdamWill:
> Hi Adam
> 
> This is metadata name rather than device node name. For example: env hostname: node1 mdadm -CR vol0 -l0 -n2 /dev/loop0 /dev/loop1 (mdadm: array /dev/md/vol0 started.) mdadm -CR vol0 -l0 -n2 /dev/loop0 /dev/loop1 --name node1 (mdadm: array /dev/md/vol0 started.)

I traced the connections out in https://bugzilla.redhat.com/show_bug.cgi?id=2325906#c19 . IMBW, but it seems to me that this *is* ultimately (at least sometimes) the name that backs the `MD_DEVNAME` property (and hence also causes a /dev/md/(thatname) symlink to be created). In your test scenario, I suspect you're running into the "we strip the hostname *if the current hostname is the same*" bit. We often won't hit that path in the installer because there is no guarantee the hostname of the installer environment is the same as the hostname of the environment that created the array (in fact it's very likely not to be).

> Hmm, I'm confused here. We always think an array already exists when talking about assemble mode. Why do you see assemble mode can create an array?

I don't mean that assemble mode creates the *array*, I mean that it can potentially create a device node (if udev is not in use on the system) or symlink (in the case you gave, if I understand it correctly) under `/dev`. i.e. it's not safe to assume, for a command like `mdadm --assemble /dev/foo ...`, that `/dev/foo` *always* already exists. It may or may not. I'm saying that if it *does* already exist, we should not apply the 'is the name valid' check, but if it does *not* yet exist, maybe we should.

XiaoNi87:
> > Hi Adam
> > This is metadata name rather than device node name. For example: env hostname: node1 mdadm -CR vol0 -l0 -n2 /dev/loop0 /dev/loop1 (mdadm: array /dev/md/vol0 started.) mdadm -CR vol0 -l0 -n2 /dev/loop0 /dev/loop1 --name node1 (mdadm: array /dev/md/vol0 started.)
> 
> I traced the connections out in https://bugzilla.redhat.com/show_bug.cgi?id=2325906#c19 . IMBW, but it seems to me that this _is_ ultimately (at least sometimes) the name that backs the `MD_DEVNAME` property (and hence also causes a /dev/md/(thatname) symlink to be created). In your test scenario, I suspect you're running into the "we strip the hostname _if the current hostname is the same_" bit. We often won't hit that path in the installer because there is no guarantee the hostname of the installer environment is the same as the hostname of the environment that created the array (in fact it's very likely not to be).

Hmm, from my test cases, I want to say mdadm doesn't create a device node /dev/md/something:something automatically if it doesn't specify /dev/md/something:something when creating an array. 

> 
> > Hmm, I'm confused here. We always think an array already exists when talking about assemble mode. Why do you see assemble mode can create an array?
> 
> I don't mean that assemble mode creates the _array_, I mean that it can potentially create a device node (if udev is not in use on the system) or symlink (in the case you gave, if I understand it correctly) under `/dev`. i.e. it's not safe to assume, for a command like `mdadm --assemble /dev/md/foo ...`, that `/dev/md/foo` _always_ already exists, although it does in the case I'm trying to fix. It may or may not. I'm saying that if it _does_ already exist, we should not apply the 'is the name valid' check, but if it does _not_ yet exist, maybe we should.

I understand you now. But in https://bugzilla.redhat.com/show_bug.cgi?id=2325906, the device node doesn't exist, right? https://bugzilla.redhat.com/show_bug.cgi?id=2325906#c14 says the array is stopped and then wants to re-start. (even if I don't know why it should do this way). 

And from my report issue, our customer wants to assemble the old arrays that devname has ':' where the /dev/md/a:b doesn't exist

AdamWill:
> Hmm, from my test cases, I want to say mdadm doesn't create a device node /dev/md/something:something automatically if it doesn't specify /dev/md/something:something when creating an array.

Nope, it definitely can. Precise reproducer:

1. Set up a VM with two entirely blank disks
2. Boot Fedora 41 Workstation live
3. Go to Installation Destination, select both disks as install targets, select Custom install
4. On the custom install screen, change the dropdown from btrfs to 'standard partition' and hit the 'create them for me' link
5. Select the root partition and change it from ext4 to 'RAID', confirm the change, and complete partitioning
6. Go back to the hub, set a root password, run the install
7. Check the mdadm commands that were run in /tmp/*.log . Nothing will have a `:` in it. The create command will look like this: `mdadm --create /dev/md/root --run --level=raid1 --raid-devices=2 --metadata=default --bitmap=internal /dev/vda3 /dev/vdb1`
8. Do `ls /dev/md*` and you'll see there's a `/dev/md/root` symlink to `/dev/md127` (or similar)
9. Reboot back to the same live image, do not run the installer
10. Do `ls /dev/md*` and there is now a `/dev/md/localhost-live:root` symlink to `/dev/md127`. `udevadm info /dev/md127` shows `MD_DEVNAME=localhost-live:root`.

Why is this the case? I don't know, honestly! It's especially odd since the hostname actually *is* the same between the two boots (it's `localhost-live` in both cases). Very possibly there's another bug in mdadm here. But it's reproducible consistently for me.

AdamWill:
That did make me wonder "what's creating these device nodes during boot?", and AFAICT the answer is "udev rules in the mdadm package". With udev debug logging enabled, it looks like these are all the mdadm commands it runs on boot which lead to the above situation:
```
Apr 03 16:13:14 fedora (udev-worker)[1187]: vda3: /usr/lib/udev/rules.d/64-md-raid-assembly.rules:51 Importing properties from results of '/usr/sbin/mdadm --incremental --export /dev/vda3 --offroot /dev/disk/by-diskseq/3-part3 /dev/disk/by-path/pci-0000:04:00.0-part3 /dev/disk/by-partuuid/cc1c1a64-acc8-4a3c-96dd-bf3bfea64ace /dev/disk/by-path/pci-0000:04:00.0-part/by-partuuid/cc1c1a64-acc8-4a3c-96dd-bf3bfea64ace /dev/disk/by-path/virtio-pci-0000:04:00.0-part3 /dev/disk/by-path/pci-0000:04:00.0-part/by-partnum/3'
Apr 03 16:13:14 fedora (udev-worker)[1187]: vda3: Starting '/usr/sbin/mdadm --incremental --export /dev/vda3 --offroot /dev/disk/by-diskseq/3-part3 /dev/disk/by-path/pci-0000:04:00.0-part3 /dev/disk/by-partuuid/cc1c1a64-acc8-4a3c-96dd-bf3bfea64ace /dev/disk/by-path/pci-0000:04:00.0-part/by-partuuid/cc1c1a64-acc8-4a3c-96dd-bf3bfea64ace /dev/disk/by-path/virtio-pci-0000:04:00.0-part3 /dev/disk/by-path/pci-0000:04:00.0-part/by-partnum/3'
Apr 03 16:13:14 fedora (udev-worker)[1174]: vdb1: /usr/lib/udev/rules.d/64-md-raid-assembly.rules:51 Importing properties from results of '/usr/sbin/mdadm --incremental --export /dev/vdb1 --offroot /dev/disk/by-path/pci-0000:08:00.0-part1 /dev/disk/by-path/virtio-pci-0000:08:00.0-part1 /dev/disk/by-diskseq/4-part1 /dev/disk/by-partuuid/1331fd3a-76d9-44c8-9514-58663fc97bfa /dev/disk/by-path/pci-0000:08:00.0-part/by-partuuid/1331fd3a-76d9-44c8-9514-58663fc97bfa /dev/disk/by-path/pci-0000:08:00.0-part/by-partnum/1'
Apr 03 16:13:14 fedora (udev-worker)[1174]: vdb1: Starting '/usr/sbin/mdadm --incremental --export /dev/vdb1 --offroot /dev/disk/by-path/pci-0000:08:00.0-part1 /dev/disk/by-path/virtio-pci-0000:08:00.0-part1 /dev/disk/by-diskseq/4-part1 /dev/disk/by-partuuid/1331fd3a-76d9-44c8-9514-58663fc97bfa /dev/disk/by-path/pci-0000:08:00.0-part/by-partuuid/1331fd3a-76d9-44c8-9514-58663fc97bfa /dev/disk/by-path/pci-0000:08:00.0-part/by-partnum/1'
Apr 03 16:13:14 fedora (udev-worker)[1187]: vda3: '/usr/sbin/mdadm --incremental --export /dev/vda3 --offroot /dev/disk/by-diskseq/3-part3 /dev/disk/by-path/pci-0000:04:00.0-part3 /dev/disk/by-partuuid/cc1c1a64-acc8-4a3c-96dd-bf3bfea64ace /dev/disk/by-path/pci-0000:04:00.0-part/by-partuuid/cc1c1a64-acc8-4a3c-96dd-bf3bfea64ace /dev/disk/by-path/virtio-pci-0000:04:00.0-part3 /dev/disk/by-path/pci-0000:04:00.0-part/by-partnum/3'(out) 'MD_DEVICE=md127'
Apr 03 16:13:14 fedora (udev-worker)[1187]: vda3: '/usr/sbin/mdadm --incremental --export /dev/vda3 --offroot /dev/disk/by-diskseq/3-part3 /dev/disk/by-path/pci-0000:04:00.0-part3 /dev/disk/by-partuuid/cc1c1a64-acc8-4a3c-96dd-bf3bfea64ace /dev/disk/by-path/pci-0000:04:00.0-part/by-partuuid/cc1c1a64-acc8-4a3c-96dd-bf3bfea64ace /dev/disk/by-path/virtio-pci-0000:04:00.0-part3 /dev/disk/by-path/pci-0000:04:00.0-part/by-partnum/3'(out) 'MD_DEVNAME=localhost-live:root'
Apr 03 16:13:14 fedora (udev-worker)[1187]: vda3: '/usr/sbin/mdadm --incremental --export /dev/vda3 --offroot /dev/disk/by-diskseq/3-part3 /dev/disk/by-path/pci-0000:04:00.0-part3 /dev/disk/by-partuuid/cc1c1a64-acc8-4a3c-96dd-bf3bfea64ace /dev/disk/by-path/pci-0000:04:00.0-part/by-partuuid/cc1c1a64-acc8-4a3c-96dd-bf3bfea64ace /dev/disk/by-path/virtio-pci-0000:04:00.0-part3 /dev/disk/by-path/pci-0000:04:00.0-part/by-partnum/3'(out) 'MD_FOREIGN=yes'
Apr 03 16:13:14 fedora (udev-worker)[1187]: vda3: '/usr/sbin/mdadm --incremental --export /dev/vda3 --offroot /dev/disk/by-diskseq/3-part3 /dev/disk/by-path/pci-0000:04:00.0-part3 /dev/disk/by-partuuid/cc1c1a64-acc8-4a3c-96dd-bf3bfea64ace /dev/disk/by-path/pci-0000:04:00.0-part/by-partuuid/cc1c1a64-acc8-4a3c-96dd-bf3bfea64ace /dev/disk/by-path/virtio-pci-0000:04:00.0-part3 /dev/disk/by-path/pci-0000:04:00.0-part/by-partnum/3'(out) 'MD_STARTED=unsafe'
Apr 03 16:13:14 fedora (udev-worker)[1187]: vda3: Process '/usr/sbin/mdadm --incremental --export /dev/vda3 --offroot /dev/disk/by-diskseq/3-part3 /dev/disk/by-path/pci-0000:04:00.0-part3 /dev/disk/by-partuuid/cc1c1a64-acc8-4a3c-96dd-bf3bfea64ace /dev/disk/by-path/pci-0000:04:00.0-part/by-partuuid/cc1c1a64-acc8-4a3c-96dd-bf3bfea64ace /dev/disk/by-path/virtio-pci-0000:04:00.0-part3 /dev/disk/by-path/pci-0000:04:00.0-part/by-partnum/3' succeeded.
Apr 03 16:13:14 fedora (udev-worker)[1187]: vda3: /usr/lib/udev/rules.d/80-udisks2.rules:20 Importing properties from results of '/bin/sh -c '/sbin/mdadm --examine --export /dev/vda3 | /bin/sed s/^MD_/UDISKS_MD_MEMBER_/g''
Apr 03 16:13:14 fedora (udev-worker)[1187]: vda3: Starting '/bin/sh -c '/sbin/mdadm --examine --export /dev/vda3 | /bin/sed s/^MD_/UDISKS_MD_MEMBER_/g''
Apr 03 16:13:14 fedora (udev-worker)[1187]: vda3: '/bin/sh -c '/sbin/mdadm --examine --export /dev/vda3 | /bin/sed s/^MD_/UDISKS_MD_MEMBER_/g''(out) 'UDISKS_MD_MEMBER_LEVEL=raid1'
Apr 03 16:13:14 fedora (udev-worker)[1187]: vda3: '/bin/sh -c '/sbin/mdadm --examine --export /dev/vda3 | /bin/sed s/^MD_/UDISKS_MD_MEMBER_/g''(out) 'UDISKS_MD_MEMBER_DEVICES=2'
Apr 03 16:13:14 fedora (udev-worker)[1187]: vda3: '/bin/sh -c '/sbin/mdadm --examine --export /dev/vda3 | /bin/sed s/^MD_/UDISKS_MD_MEMBER_/g''(out) 'UDISKS_MD_MEMBER_NAME=localhost-live:root'
Apr 03 16:13:14 fedora (udev-worker)[1187]: vda3: '/bin/sh -c '/sbin/mdadm --examine --export /dev/vda3 | /bin/sed s/^MD_/UDISKS_MD_MEMBER_/g''(out) 'UDISKS_MD_MEMBER_ARRAY_SIZE=20.38GB'
Apr 03 16:13:14 fedora (udev-worker)[1187]: vda3: '/bin/sh -c '/sbin/mdadm --examine --export /dev/vda3 | /bin/sed s/^MD_/UDISKS_MD_MEMBER_/g''(out) 'UDISKS_MD_MEMBER_UUID=725d4de6:84e9294c:a148f1e4:18e4ce05'
Apr 03 16:13:14 fedora (udev-worker)[1187]: vda3: '/bin/sh -c '/sbin/mdadm --examine --export /dev/vda3 | /bin/sed s/^MD_/UDISKS_MD_MEMBER_/g''(out) 'UDISKS_MD_MEMBER_UPDATE_TIME=1743694585'
Apr 03 16:13:14 fedora (udev-worker)[1187]: vda3: '/bin/sh -c '/sbin/mdadm --examine --export /dev/vda3 | /bin/sed s/^MD_/UDISKS_MD_MEMBER_/g''(out) 'UDISKS_MD_MEMBER_DEV_UUID=3e2428a1:78046b5d:fdebb393:7a00e03d'
Apr 03 16:13:14 fedora (udev-worker)[1187]: vda3: '/bin/sh -c '/sbin/mdadm --examine --export /dev/vda3 | /bin/sed s/^MD_/UDISKS_MD_MEMBER_/g''(out) 'UDISKS_MD_MEMBER_EVENTS=214'
Apr 03 16:13:14 fedora (udev-worker)[1187]: vda3: Process '/bin/sh -c '/sbin/mdadm --examine --export /dev/vda3 | /bin/sed s/^MD_/UDISKS_MD_MEMBER_/g'' succeeded.
Apr 03 16:13:14 fedora (udev-worker)[1189]: md127: /usr/lib/udev/rules.d/63-md-raid-arrays.rules:21 Importing properties from results of '/usr/sbin/mdadm --detail --no-devices --export /dev/md127'
Apr 03 16:13:14 fedora (udev-worker)[1189]: md127: Starting '/usr/sbin/mdadm --detail --no-devices --export /dev/md127'
Apr 03 16:13:14 fedora (udev-worker)[1189]: md127: '/usr/sbin/mdadm --detail --no-devices --export /dev/md127'(out) 'MD_LEVEL=raid1'
Apr 03 16:13:14 fedora (udev-worker)[1189]: md127: '/usr/sbin/mdadm --detail --no-devices --export /dev/md127'(out) 'MD_DEVICES=2'
Apr 03 16:13:14 fedora (udev-worker)[1189]: md127: '/usr/sbin/mdadm --detail --no-devices --export /dev/md127'(out) 'MD_METADATA=1.2'
Apr 03 16:13:14 fedora (udev-worker)[1189]: md127: '/usr/sbin/mdadm --detail --no-devices --export /dev/md127'(out) 'MD_UUID=725d4de6:84e9294c:a148f1e4:18e4ce05'
Apr 03 16:13:14 fedora (udev-worker)[1189]: md127: '/usr/sbin/mdadm --detail --no-devices --export /dev/md127'(out) 'MD_DEVNAME=localhost-live:root'
Apr 03 16:13:14 fedora (udev-worker)[1189]: md127: '/usr/sbin/mdadm --detail --no-devices --export /dev/md127'(out) 'MD_NAME=localhost-live:root'
Apr 03 16:13:14 fedora (udev-worker)[1189]: md127: Process '/usr/sbin/mdadm --detail --no-devices --export /dev/md127' succeeded.
Apr 03 16:13:14 fedora (udev-worker)[1189]: md127: /usr/lib/udev/rules.d/80-udisks2.rules:22 Importing properties from results of '/bin/sh -c '/sbin/mdadm --detail --export /dev/md127 | /bin/sed s/^MD_/UDISKS_MD_/g''
Apr 03 16:13:14 fedora (udev-worker)[1189]: md127: Starting '/bin/sh -c '/sbin/mdadm --detail --export /dev/md127 | /bin/sed s/^MD_/UDISKS_MD_/g''
Apr 03 16:13:14 fedora (udev-worker)[1189]: md127: '/bin/sh -c '/sbin/mdadm --detail --export /dev/md127 | /bin/sed s/^MD_/UDISKS_MD_/g''(out) 'UDISKS_MD_LEVEL=raid1'
Apr 03 16:13:14 fedora (udev-worker)[1189]: md127: '/bin/sh -c '/sbin/mdadm --detail --export /dev/md127 | /bin/sed s/^MD_/UDISKS_MD_/g''(out) 'UDISKS_MD_DEVICES=2'
Apr 03 16:13:14 fedora (udev-worker)[1189]: md127: '/bin/sh -c '/sbin/mdadm --detail --export /dev/md127 | /bin/sed s/^MD_/UDISKS_MD_/g''(out) 'UDISKS_MD_METADATA=1.2'
Apr 03 16:13:14 fedora (udev-worker)[1189]: md127: '/bin/sh -c '/sbin/mdadm --detail --export /dev/md127 | /bin/sed s/^MD_/UDISKS_MD_/g''(out) 'UDISKS_MD_UUID=725d4de6:84e9294c:a148f1e4:18e4ce05'
Apr 03 16:13:14 fedora (udev-worker)[1189]: md127: '/bin/sh -c '/sbin/mdadm --detail --export /dev/md127 | /bin/sed s/^MD_/UDISKS_MD_/g''(out) 'UDISKS_MD_DEVNAME=localhost-live:root'
Apr 03 16:13:14 fedora (udev-worker)[1189]: md127: '/bin/sh -c '/sbin/mdadm --detail --export /dev/md127 | /bin/sed s/^MD_/UDISKS_MD_/g''(out) 'UDISKS_MD_NAME=localhost-live:root'
Apr 03 16:13:14 fedora (udev-worker)[1189]: md127: '/bin/sh -c '/sbin/mdadm --detail --export /dev/md127 | /bin/sed s/^MD_/UDISKS_MD_/g''(out) 'UDISKS_MD_DEVICE_dev_vda3_ROLE=0'
Apr 03 16:13:14 fedora (udev-worker)[1189]: md127: '/bin/sh -c '/sbin/mdadm --detail --export /dev/md127 | /bin/sed s/^MD_/UDISKS_MD_/g''(out) 'UDISKS_MD_DEVICE_dev_vda3_DEV=/dev/vda3'
Apr 03 16:13:14 fedora (udev-worker)[1189]: md127: '/bin/sh -c '/sbin/mdadm --detail --export /dev/md127 | /bin/sed s/^MD_/UDISKS_MD_/g''(out) 'UDISKS_MD_DEVICE_dev_vdb1_ROLE=1'
Apr 03 16:13:14 fedora (udev-worker)[1189]: md127: '/bin/sh -c '/sbin/mdadm --detail --export /dev/md127 | /bin/sed s/^MD_/UDISKS_MD_/g''(out) 'UDISKS_MD_DEVICE_dev_vdb1_DEV=/dev/vdb1'
Apr 03 16:13:14 fedora (udev-worker)[1189]: md127: Process '/bin/sh -c '/sbin/mdadm --detail --export /dev/md127 | /bin/sed s/^MD_/UDISKS_MD_/g'' succeeded.
Apr 03 16:13:14 fedora (udev-worker)[1191]: md127: /usr/lib/udev/rules.d/63-md-raid-arrays.rules:21 Importing properties from results of '/usr/sbin/mdadm --detail --no-devices --export /dev/md127'
Apr 03 16:13:14 fedora (udev-worker)[1191]: md127: Starting '/usr/sbin/mdadm --detail --no-devices --export /dev/md127'
Apr 03 16:13:14 fedora (udev-worker)[1191]: md127: '/usr/sbin/mdadm --detail --no-devices --export /dev/md127'(out) 'MD_LEVEL=raid1'
Apr 03 16:13:14 fedora (udev-worker)[1191]: md127: '/usr/sbin/mdadm --detail --no-devices --export /dev/md127'(out) 'MD_DEVICES=2'
Apr 03 16:13:14 fedora (udev-worker)[1191]: md127: '/usr/sbin/mdadm --detail --no-devices --export /dev/md127'(out) 'MD_METADATA=1.2'
Apr 03 16:13:14 fedora (udev-worker)[1191]: md127: '/usr/sbin/mdadm --detail --no-devices --export /dev/md127'(out) 'MD_UUID=725d4de6:84e9294c:a148f1e4:18e4ce05'
Apr 03 16:13:14 fedora (udev-worker)[1191]: md127: '/usr/sbin/mdadm --detail --no-devices --export /dev/md127'(out) 'MD_DEVNAME=localhost-live:root'
Apr 03 16:13:14 fedora (udev-worker)[1191]: md127: '/usr/sbin/mdadm --detail --no-devices --export /dev/md127'(out) 'MD_NAME=localhost-live:root'
Apr 03 16:13:14 fedora (udev-worker)[1191]: md127: Process '/usr/sbin/mdadm --detail --no-devices --export /dev/md127' succeeded.
Apr 03 16:13:14 fedora (udev-worker)[1191]: md127: /usr/lib/udev/rules.d/80-udisks2.rules:22 Importing properties from results of '/bin/sh -c '/sbin/mdadm --detail --export /dev/md127 | /bin/sed s/^MD_/UDISKS_MD_/g''
Apr 03 16:13:14 fedora (udev-worker)[1191]: md127: Starting '/bin/sh -c '/sbin/mdadm --detail --export /dev/md127 | /bin/sed s/^MD_/UDISKS_MD_/g''
Apr 03 16:13:14 fedora (udev-worker)[1191]: md127: '/bin/sh -c '/sbin/mdadm --detail --export /dev/md127 | /bin/sed s/^MD_/UDISKS_MD_/g''(out) 'UDISKS_MD_LEVEL=raid1'
Apr 03 16:13:14 fedora (udev-worker)[1191]: md127: '/bin/sh -c '/sbin/mdadm --detail --export /dev/md127 | /bin/sed s/^MD_/UDISKS_MD_/g''(out) 'UDISKS_MD_DEVICES=2'
Apr 03 16:13:14 fedora (udev-worker)[1191]: md127: '/bin/sh -c '/sbin/mdadm --detail --export /dev/md127 | /bin/sed s/^MD_/UDISKS_MD_/g''(out) 'UDISKS_MD_METADATA=1.2'
Apr 03 16:13:14 fedora (udev-worker)[1191]: md127: '/bin/sh -c '/sbin/mdadm --detail --export /dev/md127 | /bin/sed s/^MD_/UDISKS_MD_/g''(out) 'UDISKS_MD_UUID=725d4de6:84e9294c:a148f1e4:18e4ce05'
Apr 03 16:13:14 fedora (udev-worker)[1191]: md127: '/bin/sh -c '/sbin/mdadm --detail --export /dev/md127 | /bin/sed s/^MD_/UDISKS_MD_/g''(out) 'UDISKS_MD_DEVNAME=localhost-live:root'
Apr 03 16:13:14 fedora (udev-worker)[1191]: md127: '/bin/sh -c '/sbin/mdadm --detail --export /dev/md127 | /bin/sed s/^MD_/UDISKS_MD_/g''(out) 'UDISKS_MD_NAME=localhost-live:root'
Apr 03 16:13:14 fedora (udev-worker)[1191]: md127: '/bin/sh -c '/sbin/mdadm --detail --export /dev/md127 | /bin/sed s/^MD_/UDISKS_MD_/g''(out) 'UDISKS_MD_DEVICE_dev_vda3_ROLE=0'
Apr 03 16:13:14 fedora (udev-worker)[1191]: md127: '/bin/sh -c '/sbin/mdadm --detail --export /dev/md127 | /bin/sed s/^MD_/UDISKS_MD_/g''(out) 'UDISKS_MD_DEVICE_dev_vda3_DEV=/dev/vda3'
Apr 03 16:13:14 fedora (udev-worker)[1191]: md127: '/bin/sh -c '/sbin/mdadm --detail --export /dev/md127 | /bin/sed s/^MD_/UDISKS_MD_/g''(out) 'UDISKS_MD_DEVICE_dev_vdb1_ROLE=1'
Apr 03 16:13:14 fedora (udev-worker)[1191]: md127: '/bin/sh -c '/sbin/mdadm --detail --export /dev/md127 | /bin/sed s/^MD_/UDISKS_MD_/g''(out) 'UDISKS_MD_DEVICE_dev_vdb1_DEV=/dev/vdb1'
Apr 03 16:13:14 fedora (udev-worker)[1191]: md127: Process '/bin/sh -c '/sbin/mdadm --detail --export /dev/md127 | /bin/sed s/^MD_/UDISKS_MD_/g'' succeeded.
Apr 03 16:13:14 fedora (udev-worker)[1174]: vdb1: '/usr/sbin/mdadm --incremental --export /dev/vdb1 --offroot /dev/disk/by-path/pci-0000:08:00.0-part1 /dev/disk/by-path/virtio-pci-0000:08:00.0-part1 /dev/disk/by-diskseq/4-part1 /dev/disk/by-partuuid/1331fd3a-76d9-44c8-9514-58663fc97bfa /dev/disk/by-path/pci-0000:08:00.0-part/by-partuuid/1331fd3a-76d9-44c8-9514-58663fc97bfa /dev/disk/by-path/pci-0000:08:00.0-part/by-partnum/1'(out) 'MD_DEVICE=md127'
Apr 03 16:13:14 fedora (udev-worker)[1174]: vdb1: '/usr/sbin/mdadm --incremental --export /dev/vdb1 --offroot /dev/disk/by-path/pci-0000:08:00.0-part1 /dev/disk/by-path/virtio-pci-0000:08:00.0-part1 /dev/disk/by-diskseq/4-part1 /dev/disk/by-partuuid/1331fd3a-76d9-44c8-9514-58663fc97bfa /dev/disk/by-path/pci-0000:08:00.0-part/by-partuuid/1331fd3a-76d9-44c8-9514-58663fc97bfa /dev/disk/by-path/pci-0000:08:00.0-part/by-partnum/1'(out) 'MD_DEVNAME=localhost-live:root'
Apr 03 16:13:14 fedora (udev-worker)[1174]: vdb1: '/usr/sbin/mdadm --incremental --export /dev/vdb1 --offroot /dev/disk/by-path/pci-0000:08:00.0-part1 /dev/disk/by-path/virtio-pci-0000:08:00.0-part1 /dev/disk/by-diskseq/4-part1 /dev/disk/by-partuuid/1331fd3a-76d9-44c8-9514-58663fc97bfa /dev/disk/by-path/pci-0000:08:00.0-part/by-partuuid/1331fd3a-76d9-44c8-9514-58663fc97bfa /dev/disk/by-path/pci-0000:08:00.0-part/by-partnum/1'(out) 'MD_FOREIGN=yes'
Apr 03 16:13:14 fedora (udev-worker)[1174]: vdb1: '/usr/sbin/mdadm --incremental --export /dev/vdb1 --offroot /dev/disk/by-path/pci-0000:08:00.0-part1 /dev/disk/by-path/virtio-pci-0000:08:00.0-part1 /dev/disk/by-diskseq/4-part1 /dev/disk/by-partuuid/1331fd3a-76d9-44c8-9514-58663fc97bfa /dev/disk/by-path/pci-0000:08:00.0-part/by-partuuid/1331fd3a-76d9-44c8-9514-58663fc97bfa /dev/disk/by-path/pci-0000:08:00.0-part/by-partnum/1'(out) 'MD_STARTED=yes'
Apr 03 16:13:14 fedora (udev-worker)[1174]: vdb1: Process '/usr/sbin/mdadm --incremental --export /dev/vdb1 --offroot /dev/disk/by-path/pci-0000:08:00.0-part1 /dev/disk/by-path/virtio-pci-0000:08:00.0-part1 /dev/disk/by-diskseq/4-part1 /dev/disk/by-partuuid/1331fd3a-76d9-44c8-9514-58663fc97bfa /dev/disk/by-path/pci-0000:08:00.0-part/by-partuuid/1331fd3a-76d9-44c8-9514-58663fc97bfa /dev/disk/by-path/pci-0000:08:00.0-part/by-partnum/1' succeeded.
Apr 03 16:13:14 fedora (udev-worker)[1174]: vdb1: /usr/lib/udev/rules.d/80-udisks2.rules:20 Importing properties from results of '/bin/sh -c '/sbin/mdadm --examine --export /dev/vdb1 | /bin/sed s/^MD_/UDISKS_MD_MEMBER_/g''
Apr 03 16:13:14 fedora (udev-worker)[1174]: vdb1: Starting '/bin/sh -c '/sbin/mdadm --examine --export /dev/vdb1 | /bin/sed s/^MD_/UDISKS_MD_MEMBER_/g''
Apr 03 16:13:14 fedora (udev-worker)[1174]: vdb1: '/bin/sh -c '/sbin/mdadm --examine --export /dev/vdb1 | /bin/sed s/^MD_/UDISKS_MD_MEMBER_/g''(out) 'UDISKS_MD_MEMBER_LEVEL=raid1'
Apr 03 16:13:14 fedora (udev-worker)[1174]: vdb1: '/bin/sh -c '/sbin/mdadm --examine --export /dev/vdb1 | /bin/sed s/^MD_/UDISKS_MD_MEMBER_/g''(out) 'UDISKS_MD_MEMBER_DEVICES=2'
Apr 03 16:13:14 fedora (udev-worker)[1174]: vdb1: '/bin/sh -c '/sbin/mdadm --examine --export /dev/vdb1 | /bin/sed s/^MD_/UDISKS_MD_MEMBER_/g''(out) 'UDISKS_MD_MEMBER_NAME=localhost-live:root'
Apr 03 16:13:14 fedora (udev-worker)[1174]: vdb1: '/bin/sh -c '/sbin/mdadm --examine --export /dev/vdb1 | /bin/sed s/^MD_/UDISKS_MD_MEMBER_/g''(out) 'UDISKS_MD_MEMBER_ARRAY_SIZE=20.38GB'
Apr 03 16:13:14 fedora (udev-worker)[1174]: vdb1: '/bin/sh -c '/sbin/mdadm --examine --export /dev/vdb1 | /bin/sed s/^MD_/UDISKS_MD_MEMBER_/g''(out) 'UDISKS_MD_MEMBER_UUID=725d4de6:84e9294c:a148f1e4:18e4ce05'
Apr 03 16:13:14 fedora (udev-worker)[1174]: vdb1: '/bin/sh -c '/sbin/mdadm --examine --export /dev/vdb1 | /bin/sed s/^MD_/UDISKS_MD_MEMBER_/g''(out) 'UDISKS_MD_MEMBER_UPDATE_TIME=1743694585'
Apr 03 16:13:14 fedora (udev-worker)[1174]: vdb1: '/bin/sh -c '/sbin/mdadm --examine --export /dev/vdb1 | /bin/sed s/^MD_/UDISKS_MD_MEMBER_/g''(out) 'UDISKS_MD_MEMBER_DEV_UUID=87c0bbb4:afe51d3f:79700658:8d36484e'
Apr 03 16:13:14 fedora (udev-worker)[1174]: vdb1: '/bin/sh -c '/sbin/mdadm --examine --export /dev/vdb1 | /bin/sed s/^MD_/UDISKS_MD_MEMBER_/g''(out) 'UDISKS_MD_MEMBER_EVENTS=214'
Apr 03 16:13:14 fedora (udev-worker)[1174]: vdb1: Process '/bin/sh -c '/sbin/mdadm --examine --export /dev/vdb1 | /bin/sed s/^MD_/UDISKS_MD_MEMBER_/g'' succeeded.
Apr 03 16:13:15 fedora systemd[1]: mdmonitor.service - Software RAID monitoring and management was skipped because of an unmet condition check (ConditionPathExists=/etc/mdadm.conf).
```
There's a lot there, but the key thing is that as far as I can tell, udev never provides the string `localhost-live:root` as an *input*. It does read it as an *output* from various of the commands it runs - like anaconda, it seems to be interested in the `MD_DEVNAME` and `MD_NAME` properties. It looks like this is probably what creates the `/dev/md/localhost-live:root` symlink:
```
IMPORT{program}="/usr/bin/mdadm --detail --no-devices --export $devnode"
ENV{DEVTYPE}=="disk", ENV{MD_NAME}=="?*", SYMLINK+="disk/by-id/md-name-$env{MD_NAME}", OPTIONS+="string_escape=replace"
ENV{DEVTYPE}=="disk", ENV{MD_UUID}=="?*", SYMLINK+="disk/by-id/md-uuid-$env{MD_UUID}"
ENV{DEVTYPE}=="disk", ENV{MD_DEVNAME}=="?*", SYMLINK+="md/$env{MD_DEVNAME}"
```
and indeed, this is the output of that command:
```
root@localhost-live:/home/liveuser# mdadm --detail --no-devices --export /dev/md127 
MD_LEVEL=raid1
MD_DEVICES=2
MD_METADATA=1.2
MD_UUID=725d4de6:84e9294c:a148f1e4:18e4ce05
MD_DEVNAME=localhost-live:root
MD_NAME=localhost-live:root
```

mtkaczyk:
> Oh, sorry, missed something from an earlier comment:
> 
> > In Suse or Redhat it is solved this way
> 
> To be clear, I'm the Fedora QA team lead. I work for Red Hat. The installer I'm talking about is the Red Hat / Fedora installer. That's the installer that's broken by this change. See https://bugzilla.redhat.com/show_bug.cgi?id=2325906 .

Got you. Looks like I put Anaconda and Yast to one box, forgoing this difference. If this is not true for Redhat then it should be true for Yast, unless my brain is tricking me.

mtkaczyk:
@AdamWill 

After all, I decided to revert this totally.
My idea to make mdadm better guarded itself cannot be achieved in reality.

Please let me know if #165 is fixing you issue.
I will check tomorrow if it passed testing.

mtkaczyk:
@AdamWill 
Are you good to reject this change?

AdamWill:
assuming you do something like #165 , yes. but maybe wait until that is agreed?

mtkaczyk:
Sure, we can keep it open. Thanks!

Issue #161, opened 2025-03-24
Allow creating RAID0 with 0.90 metadata


sbraz:
Hi,
After upgrading `mdadm`, I recently started running into `0.90 metadata does not support layouts for RAID0` which got added in 329dfc28debb58ffe7bd1967cea00fc583139aca. 

After reading the commit and others like 14c2161edb77d7294199e8aa7daa9f9d1d0ad5d7, I still don't really understand why the following is not allowed:
```
fallocate -l1G a
fallocate -l1G b
losetup /dev/loop0 a
losetup /dev/loop1 b
mdadm --create --assume-clean --run /dev/md99 --level=raid0 --raid-devices=2 --metadata 0.90 /dev/loop0 /dev/loop1
```

Running this results in: `mdadm: 0.90 metadata does not support layouts for RAID0`.

My first question is: why can RAID5 support layouts with metadata 0.90 but RAID0 can't?

My second question is: can I use RAID0 without a layout like I used to? Isn't there an option like `--layout NULL` to allow the command to succeed? I am not running any pre-3.14 kernels or using devices of different sizes so I think I understand the risks. It's nice to have safeguards but it would also be nice to have a way to override them when one understands the risks.

mtkaczyk:
@neilbrown could you please take a look?

Thanks!

neilbrown:
You can make the --create work by adding --layout=dangerous (I think).
It isn't really dangerous as long as the devices are the same size - as they are in this case.
super0.c should really have code similar to super1.c which checks if the sizes are all the same (modulo chunk size) and should only give the error if they are not the same.

I guess someone should write a patch.

sbraz:
Hi @neilbrown,
Thanks for the reply. I should have stated in my comment that `--layout=dangerous` does not work.
My understanding is that it sets the value to `0`, not `UnSet`:
https://github.com/md-raid-utilities/mdadm/blob/4e2e208c8d3e9ba0fae88136d7c4cd0292af73b0/maps.c#L84
Therefore, we still enter this error condition:
https://github.com/md-raid-utilities/mdadm/blob/4e2e208c8d3e9ba0fae88136d7c4cd0292af73b0/super0.c#L1324-L1328
```
# mdadm --create --assume-clean --run /dev/md99 --level=raid0 --raid-devices=2 --metadata 0.90 --layout=dangerous /dev/loop0 /dev/loop1
mdadm: 0.90 metadata does not support layouts for RAID0
```

[...]

PR #164, opened 2025-04-04
Allow RAID0 to be created with v0.90 metadata #161

neilbrown:
It is not currently possible to create a RAID0 with 0.90 metadata. This is because 0.90 cannot specify the layout of RAID0 (it is traditionally ignored) and different kernels do different things with RAID0 layouts.

However it should be possible to use --layout=dangerous as that acknowledges the risk.
It also should be possible to create a RAID0 with all devices the same size because in that case all layouts are identical.

The metadata handler can only check that all devices are the same size quite late - in write_init_super().  By that time the default is currently set - set to a value that super0 cannot handle.

So this patch delays the setting of the default value and leave it for the metadata handler (or for the Build handler).

super1 selects ORIG in that case.
intel and ddf don't support non-uniform RAID0 so they don't need any change.
super0 now checks the sizes of devices if the default RAID0 layout was requested and rejects the request in they are not the same.

validiate_geometry0 now allows "dangerous" layouts for raid0.

Fixes: #161

### DIFF
--- a/Build.c
+++ b/Build.c
@@ -103,6 +103,10 @@ int Build(struct mddev_ident *ident, struct mddev_dev *devlist, struct shape *s,
 		s->chunk = 64;
 	array.chunk_size = s->chunk*1024;
 	array.layout = s->layout;
+
+	if (array.level == 0 && array.layout == UnSet)
+		/* Raid0 leaves default to metadata handler.  That is us. */
+		array.layout = RAID0_ORIG_LAYOUT;
 	if (md_set_array_info(mdfd, &array)) {
 		pr_err("md_set_array_info() failed for %s: %s\n", chosen_name, strerror(errno));
 		goto abort;
--- a/super0.c
+++ b/super0.c
@@ -837,6 +837,7 @@ struct devinfo {
 	int fd;
 	char *devname;
 	mdu_disk_info_t disk;
+	unsigned long long dev_size;
 	struct devinfo *next;
 };
 
@@ -866,6 +867,10 @@ static int add_to_super0(struct supertype *st, mdu_disk_info_t *dinfo,
 	di->devname = devname;
 	di->disk = *dinfo;
 	di->next = NULL;
+
+	if (is_fd_valid(fd))
+		get_dev_size(fd, NULL, &di->dev_size);
+
 	*dip = di;
 
 	return 0;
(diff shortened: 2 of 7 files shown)


Build.c
```diff
@@ -103,6 +103,9 @@ int Build(struct mddev_ident *ident, struct mddev_dev *devlist, struct shape *s,
 		s->chunk = 64;
 	array.chunk_size = s->chunk*1024;
 	array.layout = s->layout;
+	if (array.level == 0 && array.layout == UnSet)
```
mtkaczyk, line 106: Could you please add empty line before and after if?

super0.c
```diff
@@ -866,6 +867,8 @@ static int add_to_super0(struct supertype *st, mdu_disk_info_t *dinfo,
 	di->devname = devname;
 	di->disk = *dinfo;
 	di->next = NULL;
+	if (is_fd_valid(fd))
```
mtkaczyk, line 870: please add empty line before and after if for readability

super0.c
```diff
@@ -923,12 +926,35 @@ static int store_super0(struct supertype *st, int fd)
 	return 0;
 }
 
+static __u64 avail_size0(struct supertype *st, __u64 devsize,
```
mtkaczyk, line 929: please move function body up.
I know that we will destroy blame, but commit message you added it has nothing interesting so I think it is fine.

eventually, you can probably call:
`st->ss->avail_size`

super0.c
```diff
@@ -923,12 +926,35 @@ static int store_super0(struct supertype *st, int fd)
...
+			s2 = avail_size0(st, di2->dev_size, 0) << 9;
+			s2 /= sb->chunk_size;
+			if (s1 != s2) {
+				pr_err("Need explicit layout=dangerous to create 0.90 raid0 on non-uniform sized devices\n");
+				return 1;
+			}
+		}
+		/* looks safe */
+		sb->layout = 0;
+	}
```
mtkaczyk, line 957: please add empty line after `if`

super0.c
```diff
@@ -923,12 +926,35 @@ static int store_super0(struct supertype *st, int fd)
...
 	int rv = 0;
 	struct devinfo *di;
 
+	if (sb->level == 0 && sb->layout == UnSet) {
+		/* Without requesting a dangerous (0) layout
+		 * we can only allow this RAID0 if all devices are
+		 * the same size
+		 */
+		for (di = st->info; di; di = di->next) {
+			struct devinfo *di2 = st->info;
```
mtkaczyk, line 943: What do you think about?
```c
	if (sb->level == 0 && sb->layout == UnSet) {
		/* Without requesting a dangerous (0) layout
		 * we can only allow this RAID0 if all devices are
		 * the same size
		 */
		unsigned long long size = 0;

		for (di = st->info; di; di = di->next) {
			unsigned long long disk_size;

			disk_size = avail_size0(st, di->dev_size, 0) << 9;
			disk_size /= sb->chunk_size;

			if (size == 0) {
				size = disk_size;
				continue;
			}

			if (size != disk_size) {
				pr_err("Need explicit layout=dangerous to create 0.90 raid0 on non-uniform sized devices\n");
				return 1;
			}
		}
```

I wanted to avoid repeated initialization of `s2` and code duplication

mtkaczyk (commented):
Hi @neilbrown 
I added few comments, please look. Nothing serious, just style improvements.

Functionality LGTM.

The test will be started again after force push, I see that job failed with `connection refused`, probably unrelated.

mtkaczyk:
All green, approved!

Waiting for confirmation from reported in #161 

I added fixes tag to PR description. Issue should be automatically closed after the merge.

PR #178, opened 2025-05-07
Minorfix

XiaoNi87:


mtkaczyk (approved):

ncroxon:
This looks like it should be broken up into 3 different PR's.

XiaoNi87:
> This looks like it should be broken up into 3 different PR's.

Hi Nigel

They are all minor fixes, so I put them into one PR. If you don't reject it so much, I want to keep it as it is now.

ncroxon (approved):

PR #180, opened 2025-05-12
mdmonitor: use MAILFROM to set sendmail envelope sender address

mwilck:
Modern mail relays may reject emails with unknown envelope sender
address.

See https://bugzilla.opensuse.org/show_bug.cgi?id=1241474 for a real-world case where this caused mdmonitor emails to be rejected by a mail relay.
    
Use the MAILFROM address also as envelope sender address (by calling sendmail with the `-f` switch) to work around this issue.


mdmonitor.c
```diff
@@ -639,11 +639,19 @@ static void execute_alert_cmd(const struct event_data *data)
...
-	FILE *mp, *mdstat;
+	FILE *mp = NULL, *mdstat;
 	char buf[BUFSIZ];
 	int n;
 
-	mp = popen(Sendmail, "w");
+	if (info.mailfrom) {
+		char *cmd = NULL;
+
+		if (xasprintf(&cmd, "%s -f%s", Sendmail, info.mailfrom) > 0)
```
mtkaczyk, line 649: I don't see a reason to not used simple snprintf instead. Do I miss something?

mwilck: It requires using a fixed-size buffer on the stack. I can do that of course. I was just using a coding pattern that I'd observed in existing mdadm code.

mtkaczyk: This one might be reported as issue by static analysis, we don't need anyone to think about it.

Please rework it if it is not a big deal :)

mwilck: Right, it's not a big deal. But I wouldn't expect static checkers to complain about this code. Anyway, will do.

mtkaczyk (approved):

PR #185, opened 2025-06-23
mdadm/super1: set ctime with __cpu_to_le64

mcimerman:


mcimerman:
Actually this is probably not desired as info->array.ctime is an `unsigned int`.

Issue #76, opened 2024-08-14
current git-master (91845da) fails to build on musl


ernsteiswuerfel:
Get this on both of my musl boxes, musl 1.2.4 and musl 1.2.5:
```
[...]
gcc -Wall -Werror -Wstrict-prototypes -Wextra -Wno-unused-parameter -Wformat -Wformat-security -Werror=format-security -fstack-protector-strong -fPIE -Warray-bounds -Wimplicit-fallthrough=0 -Wformat-overflow -Wstringop-overflow -fno-strict-overflow -fno-delete-null-pointer-checks -fwrapv -O2 -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE -DSendmail=\""/usr/sbin/sendmail -t"\" -DCONFFILE=\"/etc/mdadm.conf\" -DCONFFILE2=\"/etc/mdadm/mdadm.conf\" -DMAP_DIR=\"/run/mdadm\" -DMAP_FILE=\"map\" -DMDMON_DIR=\"/run/mdadm\" -DFAILED_SLOTS_DIR=\"/run/mdadm/failed-slots\" -DNO_COROSYNC -DNO_DLM -DVERSION=\"4.3-161-g91845dab\" -DVERS_DATE="\"2024-08-13\""  -DUSE_PTHREADS -DBINDIR=\"/sbin\"  -o restripe.o -c restripe.c
restripe.c: In function 'restore_stripes':
restripe.c:758:43: error: 'off64_t' undeclared (first use in this function); did you mean 'off_ '?
  758 |                                          (off64_t)read_offset) {
      |                                           ^~~~~~~
      |                                           off_t
restripe.c:758:43: note: each undeclared identifier is reported only once for each function it appears in
restripe.c:758:51: error: expected ')' before 'read_offset'
  758 |                                          (off64_t)read_offset) {
      |                                                   ^~~~~~~~~~~
      |                                                   )
restripe.c:757:36: note: to match this '('
  757 |                                 if (lseek64(source, read_offset, 0) !=
      |                                    ^
make: *** [Makefile:232: restripe.o] Error 1
```

mtkaczyk:
Thanks!
Looks like we need to rework it. I didn't suspect that there might be a problem with availability of off64_t.
Feel free to provide a PR if you have idea how to solve it.

@ncroxon  why we need this casting? Is there issue you observed?

ncroxon:
[...]

PR #186, opened 2025-06-30
mdadm: use lseek consistently - WIP

mtkaczyk:
mdadm used both lseek and lseek64 for legacy reasons. These days, we just need to configure __USE_LARGEFILE64 macro. Fixing this issue enables musl compilation.

Add macro, and change all lseek64 to lseek. Fix style issues in these lines.

Fixes: #76

### DIFF
--- a/Grow.c
+++ b/Grow.c
@@ -4280,10 +4280,10 @@ static int grow_backup(struct mdinfo *sra,
 		bsb.magic[15] = '2';
 	for (i = 0; i < dests; i++)
 		if (part)
-			lseek64(destfd[i], destoffsets[i] +
-				__le64_to_cpu(bsb.devstart2)*512, 0);
+			lseek(destfd[i], destoffsets[i] +
+			      __le64_to_cpu(bsb.devstart2) * 512, 0);
 		else
-			lseek64(destfd[i], destoffsets[i], 0);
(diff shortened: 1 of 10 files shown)


ncroxon (approved):
Everything looks good.

mwilck:
> Everything looks good.

So you think my remarks were wrong, misguided, or pointless? That's well possible, but I'd at least have expected some response.

ncroxon:
@mwilck, I didn't see your comment prior to my review. 
I do think the removal of the print statement is correct.
Your point about one-line raid6check.c, allows the code to fit within the 80 column restriction. 
the 4096 hardcode to a constant should be in it's own patch.

ncroxon:
The patch is correct, safe, and complete for the platforms where _FILE_OFFSET_BITS 64 is supported (which covers all modern Linux/glibc environments). All error handling and type usage is preserved.

If you need this to work on very old platforms that might lack proper large file support or where lseek does not handle 64-bit offsets, additional compatibility work might be needed. But for all practical modern Linux systems, this patch change is accurate.

mwilck:
> The patch is correct, safe, and complete 

Yeah, my comments were mostly nitpicks. I didn't mean to cause trouble. I was just looking at this because @mtkaczyk had asked for review.

Thanks!

PR #189, opened 2025-08-13
mdcheck fixes

mwilck:
This set fixes a few issues I've found with the `mdcheck_start` and `mdcheck_continue` systemd units.

- `systemctl start` hangs for these units until the check has finished, or ctrl-c is pressed.
- `systemctl stop` for these units may also hang for several minutes while the `mdcheck` script is executing `sleep`
- when these units are stopped with `systemctl stop`, the actual check in the kernel continues to run. Subsequent (re)starting these units has no effect because they see the kernel check already running.
- a check that has not been finished for an entire month (perhaps due to downtime) will be restarted from position 0 rather than picked up.

### DIFF
--- a/misc/mdcheck
+++ b/misc/mdcheck
@@ -172,5 +172,6 @@ do
 		systemctl stop mdcheck_continue.timer
 		exit 0;
 	fi
-	sleep 220
+	sleep 220 &
+	wait $!
 done
--- a/misc/mdcheck
+++ b/misc/mdcheck
@@ -66,15 +66,40 @@ shift
 
 # We need a temp file occasionally...
 tmp=/var/lib/mdcheck/.md-check-$$
-trap 'rm -f "$tmp"' 0 2 3 15
+cnt=0
 
+cleanup() {
+    # We've waited, and there are still checks running.
+    # Time to stop them.
+    for i in `eval echo {1..$cnt}`
+    do
+	eval fl=\$MD_${i}_fl
+	eval sys=\$MD_${i}_sys
+	eval dev=\$MD_${i}_dev
+
+	if [ -z "$fl" ]; then continue; fi
+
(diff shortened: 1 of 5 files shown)


mtkaczyk (approved):

mwilck:
> Let's merge it one by one

Sure. I just wanted to make you aware of the other PR.

mwilck:
Also note that I applied further improvements in #190. Perhaps you want to give that code a try as well.

mwilck:
> I reviewed the patches one by one.

Which patches exactly did you apply for your test? And did you install the service files? I need to know in order to reproduce your problem.

For testing the behavior of the systemd services, you should install all changes and run `systemctl daemon-reload`. 

> What's the change about '.service' file?

The services previously had "oneshot" type. This means that "the service manager will consider the unit up after the main process exits" (systemd.service(5)), which has implications for the method that systemd will use to try and stop the service.

> kill -TERM 838574, then it runs in the loop again.

This is weird. Does `bash` behave differently on your system? What version of bash (or better, which OS) are you running? The expected behavior is that bash receives the signal while it's executing "wait". The `trap 'exit 129' 2 3 15` line should cause it to `exit` and run the `cleanup` function on signal 0 (EXIT). This is how it behaves on my system and how it should work in general, AFAICT. Here is a trace from my system, with this full set of patches applied:

```
++ cat /sys/devices/virtual/block/md126/md/sync_action
+ '[' check '!=' check ']'
+ read a rest
+ echo 0
+ any=yes
+ '[' -z yes ']'
+ wait 17594
+ sleep 220
++ exit 129
+ cleanup
...
+ logger -p daemon.info pause checking /dev/md126 at 0
+ rm -f /var/lib/mdcheck/.md-check-17533
```

XiaoNi87 (approved):
The PR looks good to me. Thanks for these patches.

PR #190, opened 2025-08-19
More mdcheck fixes: rework mdcheck service logic

mwilck:
PR on top of #189.

This PR changes the logic of the "mdcheck" tool and the related systemd services `mdcheck_start.service` and `mdcheck_continue.service`.

The current behavior is like this:

* `mdcheck` without arguments starts a RAID check on all arrays on the system, starting at position 0. This is started from `mdcheck_start.service`, started by a systemd timer once a month.
* `mdcheck --continue` looks for files `/var/lib/mdcheck/MD_UUID_$UUID`, reads the start position from them, and starts a check from that position on the array with the respective UUID. This is started from a systemd timer every night. 

In either case, `mdcheck` won't do anything if the kernel is already running a `sync_action` on a given array. The check runs for a given period of time (default 6h) and saves the last position in the `MD_UUID` file, to be taken up when `mdcheck --continue` is called next time. When the entire array has been checked, the `MD_UUID_` file is deleted. When all checks are finished, `mdcheck_continue.timer` is stopped, to be restarted when `mdcheck_start.timer` expires next time.

Before the recent commit 8aa4ea9 ("systemd: start mdcheck_continue.timer before mdcheck_start.timer"), this could lead to a race condition when the check for a given array didn't complete throughout the month. `mdcheck_start.service` would start and reset the check position to 0 before `mdcheck_continue.service` could pick up at the last saved position. 8aa4ea9 works around this by starting mdcheck_continue.service a few minutes before mdcheck_start.timer.

Yet the general problem still exists: both services trigger checks on the kernel's part which they can only passively monitor. So if a user plays with the timer settings (which he is in his rights to do), another similar race might happen.

This patch set changes the behavior as follows:

Only `mdcheck_continue.service` actually starts and stops kernel-based sync actions. This service will continue at the saved start position if an `MD_UUID*` file exists, or start a new check at position 0 otherwise. Starting at 0 can be inhibited by creating a file `/var/lib/mdcheck/Checked_$UUID`. These files will be created by `mdcheck` when it finishes checking a given array. Thus future invocations of `mdcheck_continue.service` will not restart the check on this array.

`mdcheck_start.service` runs `mdcheck --restart`, which simply removes all `Checked_*` markers from `/var/lib/mdcheck`, so that the next invocation of `mdcheck_continue.service` will start new checks on all arrays which don't have already running checks.

The general behavior of the systemd timers and services is like before, but the mentioned race condition is avoided, even if the user modifies the timer settings arbitrarily.

This set slightly changes the behavior of the `mdcheck` script. Without `--continue`, it will still start checks on all array, but unlike before it will skip arrays for with a `Checked_` marker exists. To avoid that, run `mdcheck --restart` before `mdcheck`.

More details in the commit description of the patch "mdcheck: simplify start / continue logic and add "--restart" logic".

### DIFF
--- a/misc/mdcheck
+++ b/misc/mdcheck
@@ -129,7 +129,7 @@ do
 		logger -p daemon.info mdcheck continue checking $dev from $start
 	fi
 
-	cnt=$[cnt+1]
+	: "$((cnt+=1))"
 	eval MD_${cnt}_fl=\$fl
 	eval MD_${cnt}_sys=\$sys
 	eval MD_${cnt}_dev=\$dev
--- a/misc/mdcheck
+++ b/misc/mdcheck
@@ -37,6 +37,16 @@
 # Use "mdcheck --restart" to remove these markers and re-enable checking
 # all arrays.
 
+# If the script is run from systemd, simply write to the journal on stderr.
+# Otherwise, use logger.
+log() {
+    if [[ "$INVOCATION_ID" ]]; then
+	    echo "$@" >&2
+    else
+	    logger -p daemon.info "$@"
+    fi
+}
+
 # get device name from sysfs
 devname() {
     local dev
@@ -82,7 +92,7 @@ if [ "$cont" = yes ]; then
 		exit 1
 	fi
 elif [ "$restart" = yes ]; then
-	echo 'Re-enabling array checks for all arrays' >&2
+	log 'Re-enabling array checks for all arrays'
 	rm -f /var/lib/mdcheck/Checked_*
 	exit $?
 fi
@@ -110,7 +120,7 @@ cleanup() {
 	fi
 	echo idle > $sys/md/sync_action
 	cat $sys/md/sync_min > $fl
-	logger -p daemon.info pause checking $dev at `cat $fl`
+	log pause checking $dev at `cat $fl`
     done
     rm -f "$tmp"
 }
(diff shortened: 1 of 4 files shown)


mwilck:
As I noted above, checkpatch is complaining about the length of this line in the commit description
```
[1] https://stackoverflow.com/questions/41081417/difference-between-a-b-and-a-b-in-bash
```
I have no clue how to deal with this. Please advise. Do you want me to break the line inside the URL, or remove the line altogether?

mwilck:
Rebased on top of main branch.

XiaoNi87:
> As I noted above, checkpatch is complaining about the length of this line in the commit description
> 
> ```
> [1] https://stackoverflow.com/questions/41081417/difference-between-a-b-and-a-b-in-bash
> ```
> 
> I have no clue how to deal with this. Please advise. Do you want me to break the line inside the URL, or remove the line altogether?

I encounter failures sometimes. I think we can ignore it for this problem.

mtkaczyk:
> Not sure why upstream tests don't run. @mtkaczyk hi, Mariusz, do you know the reason?

https://github.com/md-raid-utilities/mdadm/blob/main/.github/workflows/tests.yml#L5

The file modified are probably not tested anyway so the test execution is disabled. I don't think we are installing udev rules and systemd services but I could be wrong.

mwilck:
So, are we good here?

XiaoNi87:
> So, are we good here?

Hi Martin

I need to handle two bugs this week. If you're not hurry about this, I put this in to my to do list next week.

mwilck:
No hurry. I was just wondering because the last part of the conversation was about the failed CI.

mwilck:
Closing, as we don't seem to come to an agreement her.

PR #191, opened 2025-08-20
Fix typos

Gelma:
Hi, thanks a lot for the mdadm super tool!!!

Here's a stupid patch about typos! Because perfection is in details :)

Thanks a lot,
Gelma

### DIFF
--- a/Assemble.c
+++ b/Assemble.c
@@ -216,7 +216,7 @@ static int select_devices(struct mddev_dev *devlist,
 	}
 
 	/* first walk the list of devices to find a consistent set
-	 * that match the criterea, if that is possible.
+	 * that match the criteria, if that is possible.
 	 * We flag the ones we like with 'used'.
 	 */
 	for (tmpdev = devlist;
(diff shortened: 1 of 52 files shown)


XiaoNi87:
Hi Gelma

Please check the errors of checkpatch review and fix them. If you think they are not errors, please give the explanation.

mtkaczyk:
@XiaoNi87 I'm little concerned about changing legacy error messages like "superblock not recognizable". I admit this is wrong but people may rely on it. What do you think?

XiaoNi87:
> @XiaoNi87 I'm little concerned about changing legacy error messages like "superblock not recognizable". I admit this is wrong but people may rely on it. What do you think?

Hi Mariusz

you mean following change?
pr_err("no recognisable superblock on %s.\n",
pr_err("no recognizable superblock on %s.\n",

You mean some people use the wrong message to check something wrong happen, right? If so, I think it's ok. Because the mdadm command will fail, the admin should check the return value of the mdadm command, right?

Thanks
Xiao

mtkaczyk:
@XiaoNi87 @ncroxon  are you going to review this one?

XiaoNi87:
Hi all

Because long no respond, close this first.

PR #192, opened 2025-09-01
Fix --monitor --scan with relative ARRAY devnames

QRPp:
Since commit e702f392959d ("Mdmonitor: Fix segfault"), when configuration files used non-absolute ARRAY device names, commands like `mdadm --monitor --scan` failed with `mdadm: error opening devname: No such file or directory` unless run from the `/dev/md` directory.

### DIFF
--- a/mdmonitor.c
+++ b/mdmonitor.c
@@ -254,12 +254,14 @@ int Monitor(struct mddev_dev *devlist,
 				continue;
 			if (is_devname_ignore(mdlist->devname) == true)
 				continue;
-			if (!is_mddev(mdlist->devname))
-				continue;
 
 			st = xcalloc(1, sizeof *st);
 			snprintf(st->devname, MD_NAME_MAX + sizeof(DEV_MD_DIR), DEV_MD_DIR "%s",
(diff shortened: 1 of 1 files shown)


QRPp:
(Cf. also the [Debian bug 1114744](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1114744).)

mtkaczyk (approved):

mtkaczyk:
@QRPp you have my approve.
@XiaoNi87 could you look?

QRPp:
> But I still have a question. The command doesn't fail, right? It still monitors /dev/md127 which is /dev/md/vol0.

A very good catch!  Yes, the “unconfigured” arrays were auto-detected and monitored still.

My machine in question, however, via its regular software development use, routinely created, managed and tore down MD arrays.  Hence its monitoring ignored `--monitor` alerts except for the few permanent “system” arrays, configured by name in `mdadm.conf`, filtering by those very names.

The regression was probably less severe overall than it seemed, though still not entirely benign.  Thanks for helping sort it out quickly!

Issue #195, opened 2025-09-18
Memory leak in mdadm --monitor --scan


yxist:
Hello,
We reported this bug to https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1115497
We were asked to repost it here.

We noticed a memory leak in mdadm --monitor --scan on all of our newly installed Debian 13 machines.
The worst case machine accumulated about 20GB of leaks in 13 days of uptime.

After some debugging we found the issue.
At udev.c:151 in the udev_wait_for_events function:
```
if (udev_monitor_receive_device(udev_monitor))
	return UDEV_STATUS_SUCCESS; /* event detected */
```

According to libudev docs:
> On success, udev_monitor_receive_device() returns a pointer to a newly referenced device that was received via the monitor. The caller is responsible to drop this reference when done.

As you can see, the reference to the device never gets dropped.
We put together a quick patch, which seems to have fixed the issue or at least substantially reduced the amount of leaks.
[memory_leak_fix.patch](https://github.com/user-attachments/files/22407737/memory_leak_fix.patch)
We are unfamiliar with the codebase, feel free to reformat/rewrite appropriately.

Additionally for anyone with the same issue, passing MDADM_NO_UDEV=1 env to mdadm stops the leaks as well, since it bypasses the leaking codepath.

PR #197, opened 2025-09-23
udev: Fix memleak

mtkaczyk:
According to manual:
On success, udev_monitor_receive_device() returns a pointer to a newly referenced device that was received via the monitor. The caller is responsible to drop this reference when done.

Fixes:#195

### DIFF
--- a/udev.c
+++ b/udev.c
@@ -147,9 +147,16 @@ enum udev_status udev_wait_for_events(int seconds)
 	tv.tv_sec = seconds;
 	tv.tv_usec = 0;
 
-	if (select(fd + 1, &readfds, NULL, NULL, &tv) > 0 && FD_ISSET(fd, &readfds))
-		if (udev_monitor_receive_device(udev_monitor))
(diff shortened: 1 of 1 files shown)


udev.c
```diff
@@ -147,9 +147,15 @@ enum udev_status udev_wait_for_events(int seconds)
...
-		if (udev_monitor_receive_device(udev_monitor))
-			return UDEV_STATUS_SUCCESS; /* event detected */
+	if (select(fd + 1, &readfds, NULL, NULL, &tv) > 0 && FD_ISSET(fd, &readfds)) {
+		struct udev_device *dev = udev_monitor_receive_device(udev_monitor);
+
+		if (dev)
+			udev_device_unref(dev);
+		else
+			return UDEV_STATUS_ERROR;
+		return UDEV_STATUS_SUCCESS; /* event detected */
```
mwilck, line 157: Nitpick: IMO it would look better if the `return UDEV_STATUS_SUCCESS` was moved into the `if` clause.

PR #205, opened 2025-10-16
mdadm: Create array with sync del gendisk mode

XiaoNi87:
kernel patch 9e59d609763f ("md: call del_gendisk in control path") calls del_gendisk in sync way. After the patch mentioned just now, device node (/dev/md0 .e.g) will disappear after mdadm --stop command. It resolves the problem raid can be created again because raid can be created when opening device node. Then regression tests will be interrupted.

But it causes an error when assembling array which has been fixed by pr182. But upstream people think it's not right to do so. Because people will encounter error who using old mdadm version. So in kernel space, 25db5f284fb8 ("md: add legacy_async_del_gendisk mod") is used to fix this problem. The default is async mode.

We need to set sync mode when creating array to use sync mode to del gendisk.

### DIFF
--- a/mdopen.c
+++ b/mdopen.c
@@ -148,6 +148,11 @@ int create_mddev(char *dev, char *name, int trustworthy,
 	char devnm[32];
 	char cbuf[400];
 
+	if (!init_md_mod_param()) {
+		pr_err("init md module parameters fail\n");
+		return -1;
+	}
+
 	if (!udev_is_available())
 		block_udev = 0;
 
--- a/util.c
+++ b/util.c
@@ -2557,3 +2557,48 @@ bool is_file(const char *path)
 
 	return true;
 }
+
+bool set_md_mod_parameter(const char *name, const char *value)
+{
+	char path[256];
+	int fd;
+	bool ret = true;
+
+	snprintf(path, sizeof(path), "/sys/module/md_mod/parameters/%s", name);
+
+	fd = open(path, O_WRONLY);
+	if (fd < 0) {
+		pr_err("Can't open %s\n", path);
+		return false;
+	}
+
+	if (write(fd, value, strlen(value)) != (ssize_t)strlen(value)) {
+		pr_err("Failed to write to %s\n", path);
+		ret = false;
+	}
+
+	close(fd);
+	return ret;
+}
+
+/* Init kernel md_mod parameters here if needed */
+bool init_md_mod_param(void)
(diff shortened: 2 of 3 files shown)


sysfs.c
```diff
@@ -1269,3 +1269,33 @@ bool sysfs_is_libata_allow_tpm_enabled(const int verbose)
 		return true;
 	return false;
 }
+
+bool init_md_mod_param(char *buffer, int bsize)
```
mtkaczyk: please add some function desc. I see that `*buffer` is used to get current state of legacy_async_del_gendisk but it is not obvious. This is side logic I would have described.

XiaoNi87:
Hi all

I updated the pr. As designed, we'll use sync del mode since kernel 6.18. So in this patch, it only sets /sys/module/md_mod/parameters/legacy_async_del_gendisk to N. And it doesn't restore the value. So in future, mdadm users update to the new version, they can't use async del mode anymore. And in future, we'll remove these functions when kernel space removes the kernel parameter.

mdopen.c
```diff
@@ -148,6 +148,11 @@ int create_mddev(char *dev, char *name, int trustworthy,
 	char devnm[32];
 	char cbuf[400];
 
+	if (!init_md_mod_param()) {
+		pr_err("%s:%d init md module parameters fail\n", __func__, __LINE__);
```
mtkaczyk: I think you dont need __func__, __LINE__ here.

util.c
```diff
@@ -2557,3 +2557,39 @@ bool is_file(const char *path)
...
+	if (write(fd, value, strlen(value)) != (ssize_t)strlen(value)) {
+		pr_err("Failed to write to %s\n", path);
+		ret = false;
+	}
+
+	close(fd);
+	return ret;
+}
+
+bool init_md_mod_param(void)
```
mtkaczyk: Please add description for the function

mdopen.c
mtkaczyk: Please add comment why it is needed and what we are fixing here.

mtkaczyk (commented):
This require some comments and some descriptions.
For the code LGTM.

Thanks Xiao!

util.c
XiaoNi87: Done.

mtkaczyk (approved):

ncroxon (approved):

PR #212, opened 2025-10-28
github: disable self-runners

mtkaczyk:
Runners are not avaialable, disable them. There are actions to get new runners so keep a code for a while.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,7 +1,5 @@
 name: tests
 on:
-  schedule:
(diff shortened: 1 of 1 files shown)


mwilck (approved):
Ok, but tbh, it seems to me that this workflow can be removed altogether. The "prepare machine" step seems to make assumptions about the state of the runner (`vagrant snapshot restore clean_vm_v1`) which are totally intransparent.

I don't quite understand why we can't just run the test suite on loop devices on a standard GitHub runner.

PR #213, opened 2025-10-30
GitHub workflows: mdadm-test: new workflow for mdadm test suite

mwilck:
Add a workflow that runs the mdadm test suite on a GitHub-hosted runner.

Currently some tests fail, but it's not the purpose of this PR to fix them all. That should be done subsequently.

### DIFF
--- a/.github/workflows/mdadm-test.yml
+++ b/.github/workflows/mdadm-test.yml
@@ -1,0 +1,71 @@
+---
+name: mdadm test
+on:
+  push:
+    branches:
+      - test_on_push
+    paths:
+      - '*.c'
+      - '*.h'
+      - 'tests/*'
+      - 'test'
+      - '.github/workflows/mdadm-test.yml'
+  pull_request:
+    branches:
+      - test_on_pr
+    paths:
+      - '*.c'
+      - '*.h'
+      - 'tests/*'
+      - 'test'
+      - '.github/workflows/mdadm-test.yml'
+  workflow_dispatch:
+jobs:
+  mdadm_tests:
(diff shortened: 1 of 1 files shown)


mwilck:
> @mwilck could you please disable failed tests for now? :)

What would be the benefit, except that you have green badges? I would strongly prefer to keep them enabled, as a reminder that there's stuff to be fixed.

Is there any evience that these tests would not fail outside of GitHub, or that the tests themselves are broken?

mtkaczyk (approved):

XiaoNi87:
Hi all

I read your messages and notice the failed cases part. I read the new workflow, it already specified --skip-broken. So the new workflow will not run failed cases, right?

Thanks
Xiao

mwilck:
Yes, I'm using `--skip-broken`. Still, some tests fail. You can see it in the [logs](https://github.com/md-raid-utilities/mdadm/actions/runs/18938333980). Currently failing:

```
07autoassemble
07autodetect
07testreshape5
10ddf-fail-stop-readd
24raid10deadlock
24raid456deadlock
fail07autoassemble
fail10ddf-fail-stop-readd
fail24raid456deadlock
```

mwilck:
It's 6.11.0-1018-azure

That's the latest available, AFAICT.

XiaoNi87:
It's too old. Because sometimes it can't delete the device node on time. You'll see different error cases for regression tests. It's not very meaningful to run regression tests on such old kernel.

XiaoNi87:
Hi Martin

It's good to me :)

PR #222, opened 2025-12-08
super1.c: fix crash with homehost=none

mwilck:
Problem:
mdadm --assemble --update=name --homehost="<none>" --name="myraid" "/dev/md/myraid" /dev/vdb1 /dev/vdc1
> Segmentation fault (core dumped)

homehost == NULL ist tested all over the place in mdadm, but this place has apparently been overlooked.

Suggested-by: Lidong Zhong <lidong.zhong@suse.com>


mtkaczyk:
I think that there was assumption that NULL here is not an option.
LGTM, ideally it should be validated in main/parse_conf_file but if it is the one places affected, I'm ok with the fix.

mtkaczyk (approved):

PR #223, opened 2025-12-10
variable "sigterm" must be volatile sig_atomic_t

crrodriguez:
Signals handlers can only read/write to volatile sig_atomic_t variables or atomic lock-free ones.
Anything else results in undefined behaviour.


mtkaczyk:
@crrodriguez  thanks for the fix.
Could you please tell us if there is any issue it is fixing? I'm missing this part in commit description.

crrodriguez:
It is just UB if you don't do this, that's all.

PR #227, opened 2025-12-19
mdadm: check logical block size compatibility

wuguanghao3:
After the kernel started supporting the logical block size feature, it modified the array metadata, which may lead to compatibility issues. Therefore, a compatibility warning needs to be added.

Links: https://lore.kernel.org/linux-raid/909940cf-7fdf-edc1-84b5-8bfb8f75779b@huaweicloud.com/T/#t
Signed-off-by: Wu Guanghao <wuguanghao3@huawei.com>

### DIFF
--- a/Create.c
+++ b/Create.c
@@ -48,6 +48,23 @@ static int round_size_and_verify(unsigned long long *size, int chunk)
 	return 0;
 }
 
+static void check_logical_block_size_compatibility(struct mdinfo *info)
+{
+	char buf[10] = {0};
+
+	/*
+	 * The logical block size feature is only used with metadata 1.x,
+	 * so the metadata version also needs to be checked.
+	 */
+	if (sysfs_attribute_available(info, NULL, "logical_block_size") &&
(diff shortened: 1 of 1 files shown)


Create.c
```diff
@@ -48,6 +48,23 @@ static int round_size_and_verify(unsigned long long *size, int chunk)
...
+	char buf[10] = {0};
+
+	/*
+	 * The logical block size feature is only used with metadata 1.x,
+	 * so the metadata version also needs to be checked.
+	 */
+	if (sysfs_attribute_available(info, NULL, "logical_block_size") &&
+		sysfs_get_str(info, NULL, "metadata_version", buf, sizeof(buf)) > 0 &&
+		!memcmp(buf, "1.", 2)) {
+		pr_info("Arrays created by newer kernels have the logical block size feature enabled "
```
mtkaczyk: Warning here is not really valuable. I don't think people will remember, the array creation and kernel downgrade are usually not corelated,

If you see a risk, and risk is big enough to protect user then you should verify metadata and block mdadm from assembling such array with wrong kernel.

PR #228, opened 2026-01-04
mdadm: load md_mod first

XiaoNi87:
Load md_mod first before setting module parameter legacy_async_del_gendisk Everything works well if md_mod is built in kernel. If not, create and assemble will fail.

### DIFF
--- a/.github/workflows/mdadm-test.yml
+++ b/.github/workflows/mdadm-test.yml
@@ -61,7 +61,7 @@ jobs:
         continue-on-error: true
 
       - name: Collect logs of failed tests
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         with:
           name: "mdadm-failed-test-logs"
           path: /mnt/tmp/*.log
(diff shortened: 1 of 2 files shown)


mdopen.c
```diff
@@ -38,33 +38,15 @@ int create_named_array(char *devnm)
...
-		else
-			snprintf(buf, PATH_MAX - 1, "%s",
-				 "/sbin:/usr/sbin:/usr/local/sbin");
-
-		setenv("PATH", buf, 1);
-
-		if (system("modprobe md_mod") == 0)
-			fd = open(new_array_file, O_WRONLY);
-	}
 	if (fd >= 0) {
```
mtkaczyk:
```
if (!is_fd_valid(fd)){
	pr_err("Fail to open %s\n", new_array_file);
	return 0;
}

n = write(fd, devnm, strlen(devnm));
close(fd);

if (n != (int)strlen(devnm)) {
```

Looks like simpler.

XiaoNi87: > ```
> if (!is_fd_valid(fd)){
> 	pr_err("Fail to open %s\n", new_array_file);
> 	return 0;
> }
> 
> n = write(fd, devnm, strlen(devnm));
> close(fd);
> 
> if (n != (int)strlen(devnm)) {
> ```
> 
> Looks like simpler.

Hi Mariusz

The codes mentioned above don't have relationship with this problem, right? You mean this part can be improved in this way, right?

Issue #238, opened 2026-02-07
Dracut iSCSI tests hangs at boot: Spawned process '/sbin/mdadm -I /dev/sdb' is taking longer than 59s to complete


bdrung:
The Dracut iSCSI tests 70 and 71 hang on Debian unstable with mdadm 4.5. These are the log lines found on the test with a working mdadm 4.4:

```
[   xx.yyyyyy] dracut-initqueue[pid]: Scanning devices md127 for LVM volume groups
[   xx.yyyyyy] dracut-initqueue[pid]: Found volume group "dracut" using metadata type lvm2
```

After some these warnings appear instead when using mdadm 4.5:

```
[   65.132396] (udev-worker)[521]: sdb: Spawned process '/sbin/mdadm -I /dev/sdb' [920] is taking longer than 59s to complete.
[   65.133321] systemd-udevd[512]: sdb: Worker [521] processing SEQNUM=1459 is taking a long time.
[   65.133942] systemd-udevd[512]: sdc: Worker [516] processing SEQNUM=1471 is taking a long time.
```

I bisected this regression to commit https://github.com/md-raid-utilities/mdadm/commit/f2197b6b6c14af6c788c628acd1fc6d92c268c53

The Dracut test creates a stripe RAID with two disks. Here is the script used for it: https://github.com/dracut-ng/dracut-ng/blob/main/test/TEST-70-ISCSI/create-client-root.sh

Here is the full Dracut test case: https://github.com/dracut-ng/dracut-ng/tree/main/test/TEST-70-ISCSI

Corresponding bug in Dracut: https://github.com/dracut-ng/dracut-ng/issues/2199

bdrung:
To reproduce, you can run the Dracut test case in docker/podman: clone https://github.com/dracut-ng/dracut-ng and then run:
```
test/test.sh debian:sid 70
```

xtremedata:
I do apologize for adding a comment to this closed ticket, however just for a reference this explicit check of 32MB seems to be the likely reason affecting the execution time of the 'examine' command significantly for most major cloud permanent storages, from <1s to even over a minute, multiplied by the number of devices, so making it option based is a welcomed change.

Regards,

PR #240, opened 2026-02-11
super-ddf.c: the header search now use mmap

slayercat:
Commit f2197b6b6c14 ("super-ddf: optimize DDF header search for widely
used RAID controllers")  introduces a heuristic search for DDF headers,
commonly written by RAID controllers, within the last 32MB of the disk.
However, the 4096-byte block size used for the search proved to be too
slow on network drives, leading to timeouts.

By modifying to use mmap, the responsibility of efficiently fetching
data is delegated to the kernel, thereby accelerating operations in a
remote disk environment.

Fix: Issue #238

### DIFF
--- a/super-ddf.c
+++ b/super-ddf.c
@@ -33,6 +33,10 @@
 
 #include <values.h>
 #include <stddef.h>
+#include <sys/mman.h>
+#include <unistd.h>
+#include <errno.h>
+#include <string.h>
 
 /* a non-official T10 name for creation GUIDs */
 static char T10[] = "Linux-MD";
@@ -273,7 +277,7 @@ struct phys_disk {
 #define	DDF_Missing			64
 
 
-#define SEARCH_BLOCK_SIZE  4096
+#define SEARCH_BLOCK_SIZE  (1024 * 1024)
 #define SEARCH_REGION_SIZE (32 * 1024 * 1024)
 
 /* The content of the virt_section global scope */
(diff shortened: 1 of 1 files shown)


super-ddf.c
```diff
@@ -891,85 +895,80 @@ static void *load_section(int fd, struct ddf_super *super, void *buf,
...
+	if (dsize == 0) {
+		pr_err("Device %s has size 0\n",
+			devname ? devname : "unknown");
+		return -ENODEV; /* empty device / unknown size */
+	}
 
-	search_end = dsize;
-	pos = search_start;
+	/* search for min(32MB, dsize) */
+	map_len = (dsize > SEARCH_REGION_SIZE) ? (size_t)SEARCH_REGION_SIZE
```
mwilck: If the device is < 32MB, you'll search the entire device? Wouldn't it make sense to just abort the search in that case? AFAIK a disk this small just can't hold DDF meta data.

super-ddf.c
```diff
@@ -891,85 +895,80 @@ static void *load_section(int fd, struct ddf_super *super, void *buf,
...
-
-		/*Read data from the device */
-		bytes_current_read = read(fd, buffer, bytes_block_to_read);
+	/* Suggest sequential access */
+#ifdef MADV_SEQUENTIAL
+	(void)madvise(map, map_len_adj, MADV_SEQUENTIAL);
+#endif
+#ifdef MADV_WILLNEED
+	(void)madvise(map, map_len_adj, MADV_WILLNEED);
+#endif
```
mwilck: This isn't wrong, but I don't think it will buy you much.
Actually, you may _not_ need to read the entire 32MB.

mwilck (changes requested):
Looks good in general, but I have some nitpicks. However, as mentioned in the discussion in #238, I suggest adding another commit that makes this 32MB search depend on a configuration option. This sort of non-standard DDF meta data is rare, and we slow down superblock detection on non-DDF arrays by always scanning for it.

super-ddf.c
slayercat: The call to `madvise` has been removed.

slayercat:
> Reducing 250 ms to 10 ms is a good performance gain. Did you test with a hot or cold cache? For the dracut test case the cold cache scenario is the relevant one.

After dropping the cache, the initial execution time of the previous code was approximately 250 milliseconds (Elapsed time: 0.249878291 seconds). 

The current code, on its first execution, now takes approximately 210 milliseconds (Elapsed time: 0.212811640 seconds).

mtkaczyk (approved):

PR #255, opened 2026-03-11
mdadm: fix bitmap check

bkucman:
Commit e97c4e18c84 ("mdadm: ask user if bitmap is not set") adds a prompt to the user regarding bitmap enabling during creation. However, the case of creating a container, where the bitmap is not meaningful, was omitted.

This may mislead the user, as attempting to enable bitmap ends with a container creation error: "mdadm: bitmaps not meaningful with level container"

At the stage where this check is performed, the level holds the UnSet value, LEVEL_CONTAINER is set at a later stage in Create.

The fix adds a check if the level is UnSet and set bitmap to BitmapNone, because this is the only acceptable value for the container. This check also covers the case when the level is not given during volume creation, as it makes no sense to ask for a bitmap without knowledge about level.

### DIFF
--- a/mdadm.c
+++ b/mdadm.c
@@ -1575,8 +1575,10 @@ int main(int argc, char *argv[])
 		}
 
 		if (s.btype == BitmapUnknown) {
-			if (c.runstop != 1 && s.level >= 1 &&
-			    ask("To optimize recovery speed, it is recommended to enable write-intent bitmap, do you want to enable it now?"))
+			if (s.level == UnSet)
+				s.btype = BitmapNone;
+			else if (c.runstop != 1 && s.level >= 1 &&
+				 ask("To optimize recovery speed, it is recommended to enable write-intent bitmap, do you want to enable it now?"))
 				s.btype = BitmapInternal;
 			else
 				s.btype = BitmapNone;


mwilck (approved):
Looks ok, one nitpick below.

mtkaczyk:
```c
		case O(BUILD,'l'): /* set raid level*/
			if (s.level != UnSet) {
				pr_err("raid level may only be set once.  Second value is %s.\n", optarg);
				exit(2);
			}
			s.level = map_name(pers, optarg);
			if (s.level == UnSet) {
				pr_err("invalid raid level: %s\n",
					optarg);
				exit(2);
```
Could you please share the scenario it was discovered? is it regular:
```# mdadm -CR imsm -e imsm ...```


If you will add --level=container to the command what will happen?

Did you considered something like:
```
		case O(CREATE,'e'):
		if (optarg == "imsm"
                   s.level = LEVEL_CONTAINER
```
Just asking, I don't remember exactly the flows.
At least comment it missing because the fix is an workaround for a mdadm.c rutine, maybe we should just discover container earlier as proposed but it was probably never well tested so there could be new issues.

Anyway, I think that discovering container is right way to fix this rather than workarounding the cmdline.

ncroxon:
Do we have a new rewrite of this patch ?

mtkaczyk:
No feedback for  few months, closing for now.
@bkucman please reopen it once you addressed comments

PR #258, opened 2026-03-13
Add --trigger-check and --trigger-repair option to manage mode

ncroxon:
This patch adds support for the --trigger-check and --trigger-repair option in mdadm's manage mode. The new options allows users to manually trigger a check or a repair operation on an MD array by setting the sync_action to 'check' or to 'repair' via sysfs.

Changes:
- Add TriggerCheck and TriggerRepair enum value field to context
- Add --trigger-check long option and parsing
- Add --trigger-repair long option and parsing
- Implement Manage_trigger_check_repair function
- Update help text and man page documentation

Usage: mdadm --manage /dev/md0 --trigger-check
Usage: mdadm --manage /dev/md0 --trigger-repair

### DIFF
--- a/mdadm.c
+++ b/mdadm.c
@@ -1470,10 +1470,8 @@ int main(int argc, char *argv[])
 			rv = Manage_run(ident.devname, mdfd, &c);
 		if (!rv && c.runstop < 0)
 			rv = Manage_stop(ident.devname, mdfd, c.verbose, 0);
-		if (!rv && c.trigger_check)
-			rv = Manage_trigger_check_repair(ident.devname, mdfd, 0,  c.verbose);
-		if (!rv && c.trigger_repair)
-			rv = Manage_trigger_check_repair(ident.devname, mdfd, 1, c.verbose);
+		if ((!rv && c.trigger_check) || (!rv && c.trigger_repair))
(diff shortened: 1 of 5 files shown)


mdadm.c
```diff
@@ -1462,6 +1470,10 @@ int main(int argc, char *argv[])
 			rv = Manage_run(ident.devname, mdfd, &c);
 		if (!rv && c.runstop < 0)
 			rv = Manage_stop(ident.devname, mdfd, c.verbose, 0);
+		if (!rv && c.trigger_check)
+			rv = Manage_trigger_check_repair(ident.devname, mdfd, 0,  c.verbose);
+		if (!rv && c.trigger_repair)
+			rv = Manage_trigger_check_repair(ident.devname, mdfd, 1, c.verbose);
```
mtkaczyk: If trigger_check and trigger_repair are mutually exclusive is should be verified earlier. 

If i will do:
```
mdadm --manage --trigger_check --trigger_repair
```

I think that I will start check and repair afterwards. Do we want that? Shouldn't we verify current action first?

ncroxon: I have added another patch that cleans up the issue with having both parameters on the same command. It prints a message if they both are and does not execute the command.

PR #260, opened 2026-03-16
Release mdadm-4.6

XiaoNi87:


CHANGELOG.md
```diff
@@ -1,3 +1,25 @@
+# Release [mdadm-4.6](https://git.kernel.org/pub/scm/utils/mdadm/mdadm.git/log/?h=mdadm-4.6)
```
mtkaczyk, line 1: Could you please explain why we decided to propose new release so fast. We should mention the critical bugs it is fixing clearly here.

XiaoNi87: Sure. Please check the update.

mtkaczyk (approved):

mwilck:
Don't we want to include #255?

XiaoNi87:
> Don't we want to include #255?

Hi Martin

After reviewing #255, it's not an urgent problem. So if you don't reject, I want to do the release without waiting #255.

mwilck (approved):
OK then, let's go.

PR #262, opened 2026-04-16
Fix typo in lockless bitmap warning

mwilck:
cc: @YuKuai-huawei


mtkaczyk:
@mwilck please add commit description to pass checkpatch. Thanks!

mtkaczyk (approved):

mwilck:
> @mwilck please add commit description to pass checkpatch. Thanks!

I thought this was one of the rare cases of a patch that doesn't need a description. But ok.

mtkaczyk:
> I thought this was one of the rare cases of a patch that doesn't need a description. But ok.

We can talk about that more. I have no strong opinion simply kernel style habit, if we would like to add some flexibility here,

PR #270, opened 2026-05-29
Avoid int overflow after year 2038

bmwiedemann:
Avoid int overflow after year 2038, so it can work until year 2106. Note: dates before 1970 will stop working.

See https://en.wikipedia.org/wiki/Year_2038_problem

This patch was done while reviewing potential year-2038 issues in openSUSE.

### DIFF
--- a/util.c
+++ b/util.c
@@ -634,7 +634,7 @@ int check_ext2(int fd, char *name)
 	if (sb[56] != 0x53 || sb[57] != 0xef)
 		return 0;
 
-	mtime = sb[44]|(sb[45]|(sb[46]|sb[47]<<8)<<8)<<8;
+	mtime = sb[44]|(sb[45]|(sb[46]|(unsigned int)sb[47]<<8)<<8)<<8;
 	bsize = sb[24]|(sb[25]|(sb[26]|sb[27]<<8)<<8)<<8;
 	size = sb[4]|(sb[5]|(sb[6]|sb[7]<<8)<<8)<<8;
 	size <<= bsize;


util.c
```diff
@@ -634,7 +634,7 @@ int check_ext2(int fd, char *name)
 	if (sb[56] != 0x53 || sb[57] != 0xef)
 		return 0;
 
-	mtime = sb[44]|(sb[45]|(sb[46]|sb[47]<<8)<<8)<<8;
+	mtime = sb[44]|(sb[45]|(sb[46]|(unsigned int)sb[47]<<8)<<8)<<8;
```
mwilck, line 637: This needs more explanation in the commit message. AFAIU this is effective because it overrides the [standard integral promotion in C](https://stackoverflow.com/questions/17489696/bitwise-shift-promotes-unsigned-char-to-int), which counter-intuitively promotes `unsigned char` to `int` if, as usual, the size of the latter type is larger than the size of the former.

mwilck (changes requested):
Good, but I'd appreciate a more verbose commit message. See comment.

util.c
mwilck: Extending this with explanation from claude:

 - sb is unsigned char[], so each element is unsigned char
 - C's integer promotion rules **promote unsigned char to int** before any arithmetic
 - All intermediate computations are therefore signed int
 - For timestamps after 2038-01-19 03:14:07 UTC, sb[47] >= 0x80 (the MSB has its high bit set)
 - Eventually this means a 1 is shifted into bit 31 of a signed int — this is undefined behavior in C (pre-C23), because left-shifting into the sign bit of a signed integer is UB
 - On every real implementation, **this produces a negative int**
 - When assigned to a 64-bit time_t (the type on all modern 64-bit Linux), the **negative int is sign-extended to a negative time_t**, representing a date before the Unix epoch — completely wrong

mtkaczyk:
@bmwiedemann could you please take a look?

mwilck (approved):
This has been stale for too long. My comment was only about the commit message. The change itself is looking good.

@mtkaczyk, please consider merging it.

mtkaczyk:
@mwilck
This is needed to print info that we have found ext2 with this mtime. For me, we can just print that ext2 has been found and end here. There are better tools to inspect superblock than that.

I would prefer to remove it instead, what do you think?

mwilck:
> I would prefer to remove it instead, what do you think?

Fair. Not sure why we print the mtime in the first place. I guess this was the result of a bulk code scan that @bmwiedemann did over the entire openSUSE code base.

So yeah, I'm fine with just removing the code that prints the mtime.

mtkaczyk:
Close in flavour of #298

PR #271, opened 2026-06-01
clustermd_tests: control speed limits on both nodes

Damenly:
Before this commit, record_system_speed_limit() in clustermd_tests/func.sh is never called, but restore_system_speed_limit() is always called. Which means the original system speed limits cannot be restored.

The second issue is that clustermd_tests/func.sh:control_system_speed_limit() only sets local machine's speed_limit_min and speed_limit_max but remote node won't be influenced. This causes some cases like 03r1_switch-recovery to fail because the expected operation may have been done before check.

Here introduce array nodes_test_speed_limit_min and nodes_test_speed_limit_max to set raid sync speed limits for NODE1 and NODE2. nodes_system_speed_limit_min and nodes_system_speed_limit_max are for recording origin values.

NOTE: test_speed_limit_min and test_speed_limit_max are not used by clustermd_tests anymore. Many clustermd tests require slow speed on NODE1 but faster speed on NODE2 so nodes_test_speed_limit_min and nodes_test_speed_limit_max are necessary.

Assisted-by: Codex:gpt-5.5
Signed-off-by: Su Yue <glass.su@suse.com>

### DIFF
--- a/clustermd_tests/func.sh
+++ b/clustermd_tests/func.sh
@@ -169,20 +169,58 @@ stop_md()
 }
 
 record_system_speed_limit() {
-	system_speed_limit_max=`cat /proc/sys/dev/raid/speed_limit_max`
-	system_speed_limit_min=`cat /proc/sys/dev/raid/speed_limit_min`
+	local ip
+	local idx=0
+
+	for ip in $NODE1 $NODE2
+	do
+		nodes_system_speed_limit_min[$idx]=$(ssh $ip "cat /proc/sys/dev/raid/speed_limit_min")
+		nodes_system_speed_limit_max[$idx]=$(ssh $ip "cat /proc/sys/dev/raid/speed_limit_max")
+		idx=$[idx+1]
+	done
 }
 
 # To avoid sync action finishes before checking it, it needs to limit
 # the sync speed
 control_system_speed_limit() {
-	echo $test_speed_limit_min > /proc/sys/dev/raid/speed_limit_min
-	echo $test_speed_limit_max > /proc/sys/dev/raid/speed_limit_max
+	local ip
+	local idx=0
+
+	for ip in $NODE1 $NODE2
+	do
+		ssh $ip "echo ${nodes_test_speed_limit_min[$idx]} > /proc/sys/dev/raid/speed_limit_min"
+		ssh $ip "echo ${nodes_test_speed_limit_max[$idx]} > /proc/sys/dev/raid/speed_limit_max"
+		idx=$[idx+1]
+	done
 }
 
 restore_system_speed_limit() {
-	echo $system_speed_limit_min > /proc/sys/dev/raid/speed_limit_max
-	echo $system_speed_limit_max > /proc/sys/dev/raid/speed_limit_max
+	local ip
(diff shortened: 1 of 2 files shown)


Damenly:
@mwilck @mtkaczyk @XiaoNi87 Could you review this PR?

mtkaczyk:
Hi @Damenly 
Give us more few days, please.
Thanks!

clustermd_tests/func.sh
```diff
@@ -169,20 +169,58 @@ stop_md()
 }
 
 record_system_speed_limit() {
-	system_speed_limit_max=`cat /proc/sys/dev/raid/speed_limit_max`
-	system_speed_limit_min=`cat /proc/sys/dev/raid/speed_limit_min`
```
mtkaczyk, line 173: Did you tested in against tests executed in regular environment (not clustered)?

You are removing the code that is caching the actual device state so I think that all not-clustered executions are affected. Am I right?

mtkaczyk (commented):
Hey @Damenly 
One comment but i think quite important :)

clustermd_tests/func.sh
Damenly: I just tested `./test --testdir=tests` and confirmed.

https://github.com/md-raid-utilities/mdadm/blob/2accfdfccd5170ccfda8385daebf0bcd60ebdb25/test#L233

clustermd_tests/func.sh only affects clustermd_tests. For regular tests, tests/func.sh has the same function.
It looks weird to me that running clustermd_tests in a non-clustered environment...

mtkaczyk: Ok I see now. Sorry I missed the directory tree in github view.

Very well, LGTM then. Thanks!
@mwilck could you please review?

mtkaczyk (approved):

clustermd_tests/func.sh
```diff
@@ -169,20 +169,58 @@ stop_md()
...
+	local ip
+	local idx=0
+
+	for ip in $NODE1 $NODE2
+	do
+		nodes_system_speed_limit_min[$idx]=$(ssh $ip "cat /proc/sys/dev/raid/speed_limit_min")
+		nodes_system_speed_limit_max[$idx]=$(ssh $ip "cat /proc/sys/dev/raid/speed_limit_max")
+		idx=$[idx+1]
```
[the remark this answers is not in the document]
Damenly: It's unnecessary. ssh connections are already checked by check_ssh() in check_env() before 
record_system_speed_limit().

clustermd_tests/func.sh
```diff
@@ -169,20 +169,58 @@ stop_md()
...
+	local max
+
+	for ip in $NODE1 $NODE2
+	do
+		min=${nodes_system_speed_limit_min[$idx]}
+		max=${nodes_system_speed_limit_max[$idx]}
+
+		if [ -z "$min" ]
+		then
+			echo "$ip: speed_limit_min was not recorded, skip restoring." >> $logdir/$logfile
```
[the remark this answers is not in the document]
Damenly: Right. It should be $targetdir/log.

PR #274, opened 2026-06-16
mdadm: Fix memory leak in Detail

cenhuil:
The supertype 'st' allocated in Detail.c is not freed properly in the error paths. Move the free_super call to the end of the function to ensure 'st' is always freed, preventing memory leaks.


mtkaczyk (approved):
Looks safe, I checked imsm, super0 and super1 free_super and I don't see problems.

mwilck (approved):
LGTM

PR #276, opened 2026-06-18
platform-intel: handling multiple domains

gardzinx:
In upcoming change, VMD will be working with drives connected to multiple domains (default "domain" and new "domain1" as of writing this).
Currently mdadm will not see those drives as behind VMD and will list them in a separate driver incorrectly.

VMD drives behind new domain need to be handled the same way as the drives behind old domain - detect if new domain exists and detect drives on its path for the controller.

This new domain might not be implemented yet but this patch should be backwards compatible.

Example output before this patch, ALL drives are set to be behind VMD:
```
Example output on platform:
# mdadm --detail-p
mdadm: imsm capabilities not found for controller: /sys/devices/pci0000:00/0000:00:0e.0/pci10000:e1/10000:80:17.0 (type SATA)
       Platform : Intel(R) Rapid Storage Technology enterprise
    RAID Levels : raid0 raid1 raid5 raid10
    Chunk Sizes : 4k 8k 16k 32k 64k 128k
    2TB volumes : supported
      2TB disks : supported
      Max Disks : 12
    Max Volumes : 2 per array, 12 per platform
 3rd party NVMe : supported
    NVMe Device : /dev/nvme0n1 (PHAC112301EF3P8AGN)
                    Encryption(Ability|Status): None|Unencrypted
    NVMe Device : /dev/nvme6n1 (PHLJ911200MT1P0FGN)
                    Encryption(Ability|Status): None|Unencrypted
    NVMe Device : /dev/nvme2n1 (BTLN8430055B1P6AGN)
                    Encryption(Ability|Status): None|Unencrypted

       Platform : Intel(R) Rapid Storage Technology
        Version : 22.0.0.5851
        License : Premium
    RAID Levels : raid0 raid1 raid5 raid10
    Chunk Sizes : 4k 8k 16k 32k 64k 128k
    2TB volumes : supported
      2TB disks : supported
      Max Disks : 32
    Max Volumes : 2 per array, 4 per controller
 3rd party NVMe : supported
 I/O Controller : /sys/devices/pci0000:00/0000:00:0e.0 (VMD)
 NVMe under VMD : /dev/nvme3n1 (S7YDNS0L101881W)
                    Encryption(Ability|Status): SED|Unencrypted
 NVMe under VMD : /dev/nvme5n1 (BTLF731607DC1P0GGN)
                    Encryption(Ability|Status): None|Unencrypted
 NVMe under VMD : /dev/nvme1n1 (253925805793)
                    Encryption(Ability|Status): SED|Unencrypted
```
On same setup after the patch:
```
# mdadm --detail-p
       Platform : Intel(R) Rapid Storage Technology
        Version : 22.0.0.5851
    RAID Levels : raid0 raid1 raid5 raid10
    Chunk Sizes : 4k 8k 16k 32k 64k 128k
    2TB volumes : supported
      2TB disks : supported
      Max Disks : 32
    Max Volumes : 2 per array, 4 per controller
 I/O Controller : /sys/devices/pci0000:00/0000:00:0e.0/pci10000:e1/10000:80:17.0 (SATA VMD)
          Port2 : /dev/sdb (BTWL309501EQ120LGN)
                    Encryption(Ability|Status): Other|Unencrypted
          Port0 : - no device attached -
          Port1 : - no device attached -
          Port3 : - no device attached -

       Platform : Intel(R) Rapid Storage Technology
        Version : 22.0.0.5851
        License : Premium
    RAID Levels : raid0 raid1 raid5 raid10
    Chunk Sizes : 4k 8k 16k 32k 64k 128k
    2TB volumes : supported
      2TB disks : supported
      Max Disks : 32
    Max Volumes : 2 per array, 4 per controller
 3rd party NVMe : supported
 I/O Controller : /sys/devices/pci0000:00/0000:00:0e.0 (VMD)
         Domain : /sys/devices/pci0000:00/0000:00:0e.0/pci10000:e0
 NVMe under VMD : /dev/nvme3n1 (S7YDNS0L101881W)
                    Encryption(Ability|Status): SED|Unencrypted
 NVMe under VMD : /dev/nvme5n1 (BTLF731607DC1P0GGN)
                    Encryption(Ability|Status): None|Unencrypted
 NVMe under VMD : /dev/nvme1n1 (253925805793)
                    Encryption(Ability|Status): SED|Unencrypted
         Domain : /sys/devices/pci0000:00/0000:00:0e.0/pci10000:e1
 NVMe under VMD : /dev/nvme0n1 (PHAC112301EF3P8AGN)
                    Encryption(Ability|Status): None|Unencrypted
 NVMe under VMD : /dev/nvme6n1 (PHLJ911200MT1P0FGN)
                    Encryption(Ability|Status): None|Unencrypted
 NVMe under VMD : /dev/nvme2n1 (BTLN8430055B1P6AGN)
                    Encryption(Ability|Status): None|Unencrypted
 NVMe under VMD : /dev/nvme4n1 (S7YDNS0L306079J)
                    Encryption(Ability|Status): SED|Unencrypted

```


mtkaczyk:
Thanks, for the patch, here some initial nits:
- please attach example output to the commit or PR description.
- please fix review failures.


> This new PCH domain might not be implemented yet but
this patch should be backwards compatible.

Do it mean that the feature is not yet available in upstream kernel?

gardzinx:
> Thanks, for the patch, here some initial nits:
> 
>     * please attach example output to the commit or PR description.
> 
>     * please fix review failures.
> 
> 
> > This new PCH domain might not be implemented yet but
> > this patch should be backwards compatible.
> 
> Do it mean that the feature is not yet available in upstream kernel?

Yes, this feature is not yet implemented, it is going through review process at the moment.

mwilck:
> I/O Controller : /sys/devices/pci0000:00/0000:00:0e.0 (VMD)
> I/O Controller : /sys/devices/pci0000:00/0000:00:0e.0 (VMD)

Exactly the same IO controller, twice? That looks wrong. Please explain.

gardzinx:
> > I/O Controller : /sys/devices/pci0000:00/0000:00:0e.0 (VMD)
> > I/O Controller : /sys/devices/pci0000:00/0000:00:0e.0 (VMD)
> 
> Exactly the same IO controller, twice? That looks wrong. Please explain.

Yes, the same controller is handling both devices behind PCH and CPU and it was my idea to somehow group those devices. I was also considering solution that would avoid creating separate node for PCH domain and thus list all devices in one block. Should I explore more towards that end?

mwilck:
I don't fully understand what's wrong with output "before this patch". I also don't fully understand the topology. IIUC some drives are directly attached tot he CPU, and others are attached to the South Bridge. Both are handled by the same controller device, though. IMO it doesn't matter much to users how a drive is connected (is there a noticeable performance difference?). I'd probably print the controller capabilities, and below it, list all drives that this controller manages.

Just my €0.02, like I said I don't fully understand the architecture.

gardzinx:
> I don't fully understand what's wrong with output "before this patch". I also don't fully understand the topology. IIUC some drives are directly attached tot he CPU, and others are attached to the South Bridge. Both are handled by the same controller device, though. IMO it doesn't matter much to users how a drive is connected (is there a noticeable performance difference?). I'd probably print the controller capabilities, and below it, list all drives that this controller manages.
> 
> Just my €0.02, like I said I don't fully understand the architecture.

The main problem is that before this patch, mdadm doesn't list drives as behind VMD when they are connected to PCH line. Instead there is whole new driver block 'Platform : Intel(R) Rapid Storage Technology enterprise' made.

mtkaczyk:
> Yes, the same controller is handling both devices behind PCH and CPU and it was my idea to somehow group those devices. I was also considering solution that would avoid creating separate node for PCH domain and thus list all devices in one block. Should I explore more towards that end?

Yes, please. Simply, printing the same "I/O Controller:" twice looks like a bug.

>The main problem is that before this patch, mdadm doesn't list drives as behind VMD when they are connected to PCH line. Instead there is whole new driver block 'Platform : Intel(R) Rapid Storage Technology enterprise' made.

The problem with this 'Intel(R) Rapid Storage Technology enterprise' is probably limited to Intel drives that are no longer produced and sold. Please double check. For other drives, simply they are not listed.

> Yes, this feature is not yet implemented, it is going through review process at the moment.

Well, then we cannot merge the feature in madm after it is merged in kernel. Experience tells that kernel folks may change their mind very last moment. Please  notify us when it is merged.

mwilck:
> Well, then we cannot merge the feature in madm after it is merged in kernel.

Would the feature, if merged here, cause misbhehavior if it isn't merged in the kernel? If yes, which?

gardzinx:
> > Well, then we cannot merge the feature in madm after it is merged in kernel.
> 
> Would the feature, if merged here, cause misbhehavior if it isn't merged in the kernel? If yes, which?

From my testing, this change should not change anything in current behavior. There is a check if there exists path in the controller for 'domain1' (which is not yet implemented and is affecting only certain platforms). If 'domain1' exists, it will create additional vmd node for it. The rest of changes are minor tweaks for places that use hardcoded path to 'domain' so its parameter instead.

mtkaczyk:
>> Would the feature, if merged here, cause misbhehavior if it isn't merged in the kernel? If yes, which?

>From my testing, this change should not change anything in current behavior. There is a check if there exists path in the controller for 'domain1' (which is not yet implemented and is affecting only certain platforms). If 'domain1' exists, it will create additional vmd node for it. The rest of changes are minor tweaks for places that use hardcoded path to 'domain' so its parameter instead.

@mwilck 

Yes, it is expand of the functionality to follow domains not sole domain. Should have no negative consequences.

My concerns are about "domain1", maybe folks will decide to change the naming so we will end with never-used code. But still, if you think that it is safe to merge it earlier and eventually fix later- I have no strong objections just following general policy here.

gardzinx:
Pushed a change where there is no more separate vmd node for PCH to avoid controller lame duplication during print - instead I added path to additional domain if it exists as a separate field in sys_dev struct. Modified output in description to reflect the change.

platform-intel.h
```diff
@@ -216,6 +216,7 @@ enum sys_dev_type {
 struct sys_dev {
 	enum sys_dev_type type;
 	char *path;
+	char *path_pch;
 	char *pci_id;
```
mtkaczyk, line 225: Frankly speaking, this is approach is opposite to the design. In mdadm we have `sys_dev` which should describe each individual device with unique `pci_id`.  To me it appears that `domain1` pci_id is different than `domain` and you are workaround that.  Am I correct?

How we would support "domain2" in the future?  It it not extensible.

I don't look enough detailed to suggest you the correct way of implementing it but definetelly it is not something that can be accepted without good explanation.

please also avoid platform specyfic things like "pch" in a general code. It might be reused in different context in the future.

mtkaczyk (changes requested):
Hey,
see comments.

platform-intel.c
```diff
@@ -84,6 +84,11 @@ static bool imsm_orom_support_raid_disks_count_raid10(const int raid_disks)
 	return false;
 }
 
+static char *DOMAIN_NAMES[DOMAIN_COUNT] = {
```
mtkaczyk: Shoudn't it be `VMD_DOMAINS`?

platform-intel.c
```diff
@@ -276,11 +293,16 @@ struct sys_dev *find_driver_devices(const char *bus, const char *driver)
 			continue;
 
 		if (type == SYS_DEV_VMD) {
-			char vmd_path[PATH_MAX];
+			char vmd_path[PATH_MAX - 8];
```
mtkaczyk: Why you shrinked the buffer?

platform-intel.c
mtkaczyk: Also, you should include the enum:

```c
static char *DOMAIN_NAMES[] = {
  [DOMAIN] = "domain",
  [DOMAIN1] = "domain1",
}
```

platform-intel.c
```diff
@@ -1396,6 +1447,7 @@ char *vmd_domain_to_controller(struct sys_dev *hba, char *buf)
 	struct dirent *ent;
 	DIR *dir;
 	char path[PATH_MAX];
+	int domain;
```
mtkaczyk, line 1471: If you adding something new please refactir it to reversed_Christmas_tree. Thanks!

super-intel.c
```diff
@@ -7392,37 +7422,41 @@ static int __count_volumes(char *hba_path, int dpa, int verbose,
...
-			return 0;
-		do  {
-			found = 0;
-			count += count_volumes_list(devlist,
-							NULL,
-							verbose,
-							&found);
-			dprintf("found %d count: %d\n", found, count);
-		} while (found);
+			count += active_arrays_by_format("imsm", hpath, &devlist, dpa,
```
mtkaczyk: Please fix formating

mtkaczyk:
Kernel patch has been accepted:
https://lore.kernel.org/linux-pci/20260823193353.GA1246025@bhelgaas/

mtkaczyk:
Driving this mostly on slack together. We agreed that the best we can add support for second and potentially more domains is to extend the `struct sys_hba`  to contains list of paths. It is already reflected in PR.

The open question is if we would like to have another internal list implementation or it should propose the generic list interface. Strong need of adding new feature from enterprise is giving us possibility to finally add generic list implementation to mdadm.
 I would like to hear voices from @mwilck and @ncroxon on that, so guys what do you think?
There is `dlink` but I don't like it. Much better list interface implementation is here:
https://github.com/md-raid-utilities/ledmon/blob/main/src/lib/list.h

Beside that, regular review needs to happen and changes must be validated by all affected sides (Intel and Graid).
Speaking for Graid, I don't see a space for that until November, therefore it is blocked.

ncroxon:
I prefer a generic linked-list implementation (similar to ledmon's list.h) to handle multiple domain paths. 
That is a big patch set to digest from Bjorn Helgaas.

gardzinx:
Hey guys, pushed new version that is using list implementation from ledmon. I only removed the functions that were unique to ledmon, the rest is unchanged. Let me know if that is okay :)

ncroxon:
@gardzinx 

WARNING:COMMIT_MESSAGE: Missing commit description - Add an appropriate one
Error: WARNING:COMMIT_MESSAGE: Missing commit description - Add an appropriate one


ERROR:MISSING_SIGN_OFF: Missing Signed-off-by: line(s)
Error: ERROR:MISSING_SIGN_OFF: Missing Signed-off-by: line(s)

gardzinx:
> @gardzinx
> 
> WARNING:COMMIT_MESSAGE: Missing commit description - Add an appropriate one Error: WARNING:COMMIT_MESSAGE: Missing commit description - Add an appropriate one
> 
> ERROR:MISSING_SIGN_OFF: Missing Signed-off-by: line(s) Error: ERROR:MISSING_SIGN_OFF: Missing Signed-off-by: line(s)

Fixed, also pushed change to the way vmd_domain_to_controller detects the correct domain (from relying on hardcoded value to actually comparing pci_id with domain links).

super-intel.c
```diff
@@ -741,14 +741,37 @@ const char *get_sys_dev_type(enum sys_dev_type type)
...
+struct list *copy_paths(const struct list *src)
+{
+	if (src == NULL)
+		return NULL;
+
+	struct list *dst = xmalloc(sizeof(struct list));
+
+	list_init(dst, free);
+
+	char *path;
```
mtkaczyk: This should be declared on the top

PR #279, opened 2026-06-30
Udev-related fixes and cleanups

prajnoha:
This patchset cleans up udev blocking lifecycle management across Incremental, Create, and Assemble, fixing several bugs and a race condition along the way.

Problems addressed:

1. Symlink wait vs. uevent ordering race. In the Incremental container path, wait_for() ran before the change uevent was sent. udev could see the initial add event before metadata was set, skip symlink creation, and cause a ~3.6s timeout. The same misordering existed in the non-container path and in Assemble (relative to sysfs_rules_apply()).

2. Leaked udev blocking files. Several error paths in create_mddev() and Incremental_container() returned without calling udev_unblock(), leaving the blocking file on disk. Skipped container members also leaked the blocking file.

3. Redundant/spurious udev calls. udev_unblock() and sysfs_uevent() were called unconditionally even when udev was not available or when the code path never called udev_block() (e.g. a second disk arriving for an already-existing IMSM container). The container release path sent a redundant uevent that subarrays had already sent individually.

### DIFF
--- a/Create.c
+++ b/Create.c
@@ -1331,8 +1331,7 @@ int Create(struct supertype *st, struct mddev_ident *ident, int subdevs,
 		pr_err("not starting array - not enough devices.\n");
 	}
 	close(mdfd);
-	udev_unblock();
-	sysfs_uevent(&info, "change");
+	udev_ready(&info);
 	dev_policy_free(custom_pols);
 
 	return 0;
--- a/Create.c
+++ b/Create.c
@@ -523,6 +523,7 @@ int Create(struct supertype *st, struct mddev_ident *ident, int subdevs,
 	int container_fd = -1;
 	int need_mdmon = 0;
 	unsigned long long bitmapsize;
+	bool udev_blocked;
 	struct mdinfo info;
 	int did_default = 0;
 	int do_default_layout = 0;
@@ -1029,7 +1030,8 @@ int Create(struct supertype *st, struct mddev_ident *ident, int subdevs,
 
 	/* We need to create the device */
 	map_lock(&map);
-	mdfd = create_mddev(ident->devname, ident->name, LOCAL, chosen_name, 1);
+	udev_blocked = udev_is_available();
+	mdfd = create_mddev(ident->devname, ident->name, LOCAL, chosen_name, udev_blocked);
 	if (mdfd < 0) {
 		map_unlock(&map);
 		return 1;
@@ -1043,7 +1045,8 @@ int Create(struct supertype *st, struct mddev_ident *ident, int subdevs,
 		pr_err("Array name %s is in use already.\n", chosen_name);
 		close(mdfd);
 		map_unlock(&map);
-		udev_unblock();
+		if (udev_blocked)
+			udev_unblock();
 		return 1;
 	}
 
@@ -1331,13 +1334,15 @@ int Create(struct supertype *st, struct mddev_ident *ident, int subdevs,
 		pr_err("not starting array - not enough devices.\n");
 	}
 	close(mdfd);
-	udev_ready(&info);
+	if (udev_blocked)
+		udev_ready(&info);
 	dev_policy_free(custom_pols);
 
 	return 0;
 
  abort:
-	udev_unblock();
+	if (udev_blocked)
+		udev_unblock();
 	map_lock(&map);
  abort_locked:
 	map_remove(&map, fd2devnm(mdfd));
--- a/Assemble.c
+++ b/Assemble.c
@@ -2221,8 +2221,8 @@ int assemble_container_content(struct supertype *st, int mdfd,
 		set_array_assembly_status(c, result, INCR_NO, &array);
 	else {
 		set_array_assembly_status(c, result, INCR_YES, &array);
-		wait_for(chosen_name, mdfd);
 		sysfs_rules_apply(chosen_name, content, st);
+		wait_for(chosen_name, mdfd);
 	}
 
 	return err;
--- a/Incremental.c
+++ b/Incremental.c
@@ -1614,6 +1614,7 @@ static int Incremental_container(struct supertype *st, char *devname,
 		}
 
 		if (only && (!mp || strcmp(mp->devnm, only) != 0)) {
+			udev_unblock();
 			close_fd(&mdfd);
 			continue;
 		}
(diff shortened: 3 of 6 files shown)


mtkaczyk:
Hello @prajnoha 
Thanks for the proposal. I see it risky. This utility has been developed for a years and I saw a lot weird things related to udev especially with fragile external management.

I'm not saying that we are not going to merge this but it would take time before it is deeply validated. Did you tested it with exernal metadata?

@bkucman could you please test this with IMSM?

prajnoha:
> I'm not saying that we are not going to merge this but it would take time before it is deeply validated. Did you tested it with exernal metadata?

Yes, I have tested with IMSM (with `IMSM_NO_PLATFORM="1"`). Maybe I could provide the udev monitor output before and after using the patches, so you can compare.

mtkaczyk:
> Yes, I have tested with IMSM (with IMSM_NO_PLATFORM="1"). Maybe I could provide the udev monitor output before and after using the patches, so you can compare.

I'm more concerned about boot flows with OS installed on IMSM raid here is why I asked bkucman to verify (imsm is owned by graid now).

bkucman:
Yes, this change requires testing on real test case scenarios, I’ll internally plan such tests on the hardware with IMSM.

mtkaczyk:
Hey @bkucman could you please give an testing ETA here?

Create.c
```diff
@@ -1029,7 +1030,8 @@ int Create(struct supertype *st, struct mddev_ident *ident, int subdevs,
 
 	/* We need to create the device */
 	map_lock(&map);
-	mdfd = create_mddev(ident->devname, ident->name, LOCAL, chosen_name, 1);
+	udev_blocked = udev_is_available();
```
mwilck, line 1033: Instead of calling `udev_is_available` explicitly in callers (which can easily be forgotten), why not keep the call in `create_mddev()`, and add an `int *` (or `bool *`) parameter in which `create_mddev()` can store the current state?

mwilck: @mtkaczyk:

Let me add that I think the entire `udev_block()` logic is racy by design. 
We call `udev_is_available()` in many places now. Apparently the idea is that the availability of udev can change while mdadm is running (otherwise, one call to `udev_is_available()` at startup would be sufficient). But if we assume that this can happen, it can happen any time, so remembering this state in a variable is moot.

TBH, I don't  understand why we need the `udev_is_available()` check at all. Here at least. If udev was not running, creating files under `/run/mdadm` and deleting them later wouldn't do any harm. So we could just create and delete them always and skip the test if udev is running. Or what am I overlooking here?

mwilck (changes requested):
Minor remarks from my side. Sorry for the late review.

Create.c
prajnoha: > Instead of calling `udev_is_available` explicitly in callers (which can easily be forgotten), why not keep the call in `create_mddev()`, and add an `int *` (or `bool *`) parameter in which `create_mddev()` can store the current state?

Yes, sure, we can do that instead - that would do the job as well.

> Let me add that I think the entire `udev_block()` logic is racy by design. We call `udev_is_available()` in many places now. Apparently the idea is that the availability of udev can change while mdadm is running (otherwise, one call to `udev_is_available()` at startup would be sufficient). But if we assume that this can happen, it can happen any time, so remembering this state in a variable is moot.

Honestly, I'm not a big fan of using any files in the filesystem for any notification purposes, because then it's critical for us to be sure we always clean it up properly. And yes, it is usually prone to various races too. However, this is the existing mechanism we have. Surely, we could start thinking of improving it and look for better future mechanism. For now, this patchset is just a fixup/cleanup for existing one (which I came to and noticed an opportunity for immediate smaller improvements).

> TBH, I don't understand why we need the `udev_is_available()` check at all. Here at least. If udev was not running, creating files under `/run/mdadm` and deleting them later wouldn't do any harm. So we could just create and delete them always and skip the test if udev is running. Or what am I overlooking here?

For me, that would be just a matter of not executing useless code. BUT `udev_is_available` now executes `stat` for `/dev/.udev` and `/run/udev` each time, which itself seems superfluous - running it at the start should suffice. I'd double check the error paths though, where certain operation starts with "udev enabled" and then later in the process, it encounters a situation where udev is gone (like, not waiting for a signal from udev indefinitely or similar). Just in case.

Overall, sure, we can start thinking about more broader cleanups and improvements here for future changes.

mwilck:
> I'm not a big fan of using any files in the filesystem for any notification purposes

Not sure what other options we have. we could use `flock()` or the like, but it would come with increased likelihood of deadlocks.

> For me, that would be just a matter of not executing useless code.

Yes. But it's useless code one way or the other. Either we create and delete files under /run pointlessly, or we repeatedly check for udev, also pointlessly. Either operation is rather light-weight, I think.

Create.c
mtkaczyk: `is_udev_available` is a 2 years ago reworked verification that has been always here. I never tried to challenge that and I trust your judgement. Definitely, there is a log of space to rework.

mtkaczyk: With udev and the limited experience and support we always had, we (previously at  Intel) at Graid always minimized risk of regression in this area.

I'm glad to see you guys talking about this and challenging this legacy code. You have my support but we have limited validation capabilities for a while, we are still building everything.

Sorry @prajnoha for keeping you waiting for regression testing.

mwilck: > Sorry @prajnoha for keeping you waiting for regression testing.

Personally, I see no risk of regeressions here.

mtkaczyk: I'm more careful here as any mistake may lead to mdmon blocked or not stated which is critical.

prajnoha:
Hi! I have refreshed the patchset and addressed some of the points from our discussion - mainly the `udev_is_available` check and remembering the state at the beginning instead of checking it at each call. Also a few more patches added.

I'm also looking at  `udev_wait_for_events` which somehow doesn't make sense to me at the moment - it seems to be waiting for *any* single uevent for any block device? Hmm. Before, it was waiting for the whole udev queue to settle, which seemed nearer to what we need there, I think, but then it was replaced by this single uevent wait later with commit 9935cf0f64f3f1e70e7840385e9838c30487dc64. I'll try to handle this later in a separate PR when I find a more suitable solution here.

PR #280, opened 2026-07-03
mdstat_ent handling rework

mtkaczyk:
No functional changes, just a follow-up of previous actions with a goal of decentralizing mdadm.

### DIFF
--- a/Assemble.c
+++ b/Assemble.c
@@ -23,6 +23,7 @@
  */
 
 #include	"mdadm.h"
+#include	"mdstat.h"
 #include	"xmalloc.h"
 
 #include	<ctype.h>
--- a/Assemble.c
+++ b/Assemble.c
@@ -121,7 +121,7 @@ static int is_member_busy(char *metadata_version)
 			continue;
 
 		/* Skip first char - it can be '/' or '-' */
-		if (strcmp(&ent->metadata_version[10], metadata_version + 1) == 0) {
+		if (strcmp(&ent->metadata_version[MD_VER_EXT_LEN + 1], metadata_version + 1) == 0) {
 			busy = 1;
 			break;
 		}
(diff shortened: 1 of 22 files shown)


Assemble.c
```diff
@@ -120,7 +121,8 @@ static int is_member_busy(char *metadata_version)
 			continue;
 
 		/* Skip first char - it can be '/' or '-' */
-		if (strcmp(&ent->metadata_version[10], metadata_version + 1) == 0) {
+		if (strcmp(&ent->metadata_version[MD_VER_EXT_LEN + 1],
+			   metadata_version + 1) == 0) {
```
bkucman: What line length you prefere in mdadm?
If 100ch, you don't have to break it.

msg.c
bkucman:
```suggestion
		sprintf(buf, "%s%s\n", MD_VER_EXT, sra->text_version);
```

bkucman:
```suggestion
	sprintf(buf, "%s%s\n", MD_VER_EXT, sra->text_version);
```

bkucman (changes requested):
Hi, thank you for your patch.

there are still two more spot with "external:", you could also use define, unless you think that's overkill.

mdadm.c:1857
int member = e->metadata_version &&
	strncmp(e->metadata_version,
		"external:/", 10) == 0;

monitor.c:495
if (strncmp(buf, "external:-", 10) == 0) {

mtkaczyk:
> there are still two more spot with "external:", you could also use define, unless you think that's overkill.

lets proceed with this smaller rework. What you are asking is complete rework of how to set and read the version which I don't have good concept yet.

Noted this as a another thing to rework.

bkucman (approved):

PR #286, opened 2026-07-31
[PATCH] super-ddf: Prevent OOB heap reads in examine_vds() and examine_pds()

ncroxon:
In examine_vds() and examine_pds(), loop limits are driven by max_vdes and max_pdes read directly from the disk metadata structures (sb->virt and sb->phys). If these fields specify more entries than fit inside the section buffer (allocated based on virt_section_length and phys_section_length), mdadm will perform out-of-bounds heap reads during examination.

Fix this by ensuring max_vdes and max_pdes are clamped to the maximum number of entries that fit within the allocated section payload using element dereferencing (*entries) for element size calculation.


XiaoNi87 (commented):
Hi Nigel

Thanks for this patch. The function examine_vd is called by examine_vds and it has this problem also. So it's still possible to hit this problem. Can you put the codes into a helper function and fix the problem in examine_vd?

ncroxon:
> Hi Nigel
> 
> Thanks for this patch. The function examine_vd is called by examine_vds and it has this problem also. So it's still possible to hit this problem. Can you put the codes into a helper function and fix the problem in examine_vd?

Hello Xiao,
I have re-written the patch and uploaded the new version.
-Nigel

mtkaczyk (approved):

XiaoNi87 (approved):
Thanks, Nigel.

PR #287, opened 2026-08-06
[PATCH] super-ddf: fix mppe OOB read in examine_vd

ncroxon:
DDF record validation and the examine_vd() iteration bounds.

Validation of conf_rec_len / mppe in load_ddf_global() new helper ddf_max_primary_entries() to compute safe entry capacity bounds-limited examine_vd() loop using validated record size.
This prevents the phys_refnum[i] and LBA_OFFSET(sb, vc)[i] OOB reads.

### DIFF
--- a/super-ddf.c
+++ b/super-ddf.c
@@ -1121,7 +1121,23 @@ static int load_ddf_global(int fd, struct ddf_super *super, char *devname)
 	super->max_part = be16_to_cpu(super->active->max_partitions);
 	super->mppe = be16_to_cpu(super->active->max_primary_element_entries);
 	super->conf_rec_len = be16_to_cpu(super->active->config_record_len);
+	if (super->conf_rec_len * 512 <
+	    offsetof(struct vd_config, phys_refnum) +
+	    (sizeof(__u32) + sizeof(be64)) * super->mppe) {
+		pr_err("failed to load the DDF global information sections\n");
+		return 2;
+	}
 	return 0;
+}
+
+static unsigned int ddf_max_primary_entries(const struct ddf_super *ddf)
+{
+	unsigned int record_size = ddf->conf_rec_len * 512;
+
+	if (record_size < offsetof(struct vd_config, phys_refnum))
+		return 0;
+	return (record_size - offsetof(struct vd_config, phys_refnum)) /
+		(sizeof(__u32) + sizeof(be64));
 }
 
 #define DDF_UNUSED_BVD 0xff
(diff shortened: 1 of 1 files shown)


XiaoNi87:
Hi Nigel

max_count is initialized twice in the patch. But max_count is not used after the first initialization. 
unsigned int max_count = sb->mppe;
max_count = ddf_max_primary_entries(sb);

Since LBA_OFFSET(sb, vc) is computed from sb->mppe, both phys_refnum[i] and LBA_OFFSET(sb, vc)[i] must only be indexed up to sb->mppe. The current code can allow prim_count > sb->mppe if ddf_max_primary_entries(sb) returns a larger value.

So it needs to use min function to choose the smaller one, like this:
  max_count = ddf_max_primary_entries(sb);
  if (max_count > sb->mppe)
  	max_count = sb->mppe;

By the way, do you have a reproducer to reproduce this problem?

super-ddf.c
```diff
@@ -1563,17 +1577,26 @@ static void examine_vd(int n, struct ddf_super *sb, char *guid)
 
 	for (vcl = sb->conflist ; vcl ; vcl = vcl->next) {
 		unsigned int i;
+		unsigned int max_count = sb->mppe;
 		struct vd_config *vc = &vcl->conf;
+		unsigned int prim_count = be16_to_cpu(vc->prim_elmnt_count);
```
mtkaczyk, line 1582: please follow reversed Christmas tree (of course as much as possible):
```c
		struct vd_config *vc = &vcl->conf;
		unsigned int prim_count = be16_to_cpu(vc->prim_elmnt_count);
		unsigned int max_count = sb->mppe;
		unsigned int i;
```

super-ddf.c
```diff
@@ -1563,17 +1577,26 @@ static void examine_vd(int n, struct ddf_super *sb, char *guid)
...
 
 		if (!be32_eq(calc_crc(vc, crl*512), vc->crc))
 			continue;
 		if (memcmp(vc->guid, guid, DDF_GUID_LEN) != 0)
 			continue;
 
+		max_count = ddf_max_primary_entries(sb);
```
mtkaczyk, line 1589: this is initially assigned to `max_count = sb->mppe;` and it is unused.

super-ddf.c
```diff
@@ -1121,9 +1121,23 @@ static int load_ddf_global(int fd, struct ddf_super *super, char *devname)
 	super->max_part = be16_to_cpu(super->active->max_partitions);
 	super->mppe = be16_to_cpu(super->active->max_primary_element_entries);
 	super->conf_rec_len = be16_to_cpu(super->active->config_record_len);
+	if (super->conf_rec_len * 512 <
+	    offsetof(struct vd_config, phys_refnum) +
+	    (sizeof(__u32) + sizeof(be64)) * super->mppe)
+		return 2;
```
mtkaczyk, line 1127: can we have error message here? It will be executed on `load_super` path

super-ddf.c
ncroxon: fixed in V2:

super-ddf.c
ncroxon: added a pr_err message in V2.

super-ddf.c
ncroxon: fixed in V2

ncroxon:
@XiaoNi87 I fixed your reported issue in V2

PR #288, opened 2026-08-09
raid6check.c: Use batching to speed up I/O

davidgfnet:
raid6check works at one stripe (per disk) at a time, which makes it not I/O bound. By batching I/O operations we significantly improve its performance at some reasonable memory cost (picked 512MB as a reasonable number, since it won't improve much more with bigger values, and it's a reasonable amount of memory to expect from a RAID machine these days).

### DIFF
--- a/raid6check.c
+++ b/raid6check.c
@@ -30,6 +30,9 @@
 #define CHECK_PAGE_BITS (12)
 #define CHECK_PAGE_SIZE (1 << CHECK_PAGE_BITS)
 
+/* Data read per iteration, to amortise the lock, seek and read calls */
+#define CHECK_BATCH_BYTES (512ULL << 20)
+
 char const Name[] = "raid6check";
 
 enum repair {
@@ -129,8 +132,9 @@ void raid6_stats(int *disk, int *results, int raid_disks, int chunk_size)
 	}
 }
 
-int lock_stripe(struct mdinfo *info, unsigned long long start,
-		int chunk_size, int data_disks, sighandler_t *sig)
+int lock_stripes(struct mdinfo *info, unsigned long long start,
+		 unsigned long long nstripes, int chunk_size, int data_disks,
+		 sighandler_t *sig)
 {
 	int rv;
 
@@ -146,7 +150,7 @@ int lock_stripe(struct mdinfo *info, unsigned long long start,
 	}
 
 	rv = sysfs_set_num(info, NULL, "suspend_lo", start * chunk_size * data_disks);
-	rv |= sysfs_set_num(info, NULL, "suspend_hi", (start + 1) * chunk_size * data_disks);
+	rv |= sysfs_set_num(info, NULL, "suspend_hi", (start + nstripes) * chunk_size * data_disks);
 	return rv * 256;
 }
 
@@ -357,24 +361,27 @@ int check_stripes(struct mdinfo *info, int *source, unsigned long long *offsets,
 	if (!tables_ready)
 		make_tables();
 
-	if (posix_memalign((void**)&stripe_buf, 4096, raid_disks * chunk_size) != 0)
+	/* Stripes are checked one at a time but read in batches. The buffer is
+	 * device major, so each device's chunks are contiguous.
+	 */
+	unsigned long long batchf = min(length, CHECK_BATCH_BYTES /
+				 ((unsigned long long)raid_disks * chunk_size));
+	size_t device_span = batchf * chunk_size;
+
+	if (posix_memalign((void**)&stripe_buf, 4096, raid_disks * device_span) != 0)
 		exit(4);
 	block_index_for_slot += 2;
 	blocks += 2;
 	blocks_page += 2;
 
 	memset(zero, 0, chunk_size);
-	for ( i = 0 ; i < raid_disks ; i++)
-		stripes[i] = stripe_buf + i * chunk_size;
 
 	while (length > 0) {
-		/* The syndrome number of the broken disk is recorded
-		 * in 'disk[]' which allows a different broken disk for
-		 * each page.
-		 */
-		int disk[chunk_size >> CHECK_PAGE_BITS];
+		unsigned long long batchcnt = min(batchf, length);
+		size_t batch_bytes = batchcnt * chunk_size;
+		unsigned long long s;
 
-		err = lock_stripe(info, start, chunk_size, data_disks, sig);
+		err = lock_stripes(info, start, batchcnt, chunk_size, data_disks, sig);
 		if(err != 0) {
 			if (err != 2)
 				unlock_all_stripes(info, sig);
@@ -389,8 +396,9 @@ int check_stripes(struct mdinfo *info, int *source, unsigned long long *offsets,
 				err = -1;
 				goto exitCheck;
 			}
-			int read_res = read(source[i], stripes[i], chunk_size);
-			if (read_res < chunk_size) {
+			ssize_t read_res = read(source[i], stripe_buf + i * device_span,
+						batch_bytes);
+			if (read_res < (ssize_t)batch_bytes) {
 				fprintf(stderr, "Failed to read complete chunk disk %d, aborting\n", i);
 				unlock_all_stripes(info, sig);
 				err = -1;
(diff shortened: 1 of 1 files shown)


mtkaczyk:
Hey @davidgfnet 
Could you please provide some numbers form your tests? How much it improved the test time?

Thanks!

davidgfnet:
On my test setup (it's a Haswell machine with 32GiB of RAM) that has a 6 disk RAID6 setup, this results in a 3.5X speed up. The original performance is very slow (~40MiB/s across all disks!), due to the way it's implemented.
I calculated it would take me more than a week to fully test my RAID setup (it had some minor errors and I wanted to find them and fix them since one disk had some minor corruption) which is way too long. Not only that it sits at 10% CPU usage (and ~10% I/O usage)
After staring at the code I noticed that there's no CPU nor disk bottleneck but the whole syscall to lock pages and what not. I did add some more optimizations (calling mlock fewer times, using AIO to read() data) which resulted in much more performance gain (don't recall but it might have been around 10X in the end), but I wasn't sure whether you'd accept all of them, so I sent a PR for the "easiest" one of them: batch reading.

Lemme know if you need more info or changes, happy to help. The file could use a cleanup since the checker throws a few sytle errors and warnings, happy to send a PR to address those too.

mtkaczyk:
>Lemme know if you need more info or changes, happy to help. The file could use a cleanup since the checker throws a few sytle errors and warnings, happy to send a PR to address those too.

Go for it and have fun!
Please also add how it improved the performance on your setup to commit message and I think we are good.
I totally believe that this code is not optimized.

davidgfnet:
Allright there we go!
I will send a couple more PRs after this one, one to clean things up a bit and make the linter happy and another one to add AIO. Using AIO on my setup gives another 3x speedup, which makes the whole thing be 10x faster than the baseline tools. That helped me a lot, since a full check would take <24h (vs 7+ days).

mtkaczyk:
> ERROR:MISSING_SIGN_OFF: Missing Signed-off-by: line(s)
> Error: ERROR:MISSING_SIGN_OFF: Missing Signed-off-by: line(s)

Please fix this.

davidgfnet:
Sorry about that!

raid6check.c
```diff
@@ -357,24 +361,27 @@ int check_stripes(struct mdinfo *info, int *source, unsigned long long *offsets,
...
 
-	if (posix_memalign((void**)&stripe_buf, 4096, raid_disks * chunk_size) != 0)
+	/* Stripes are checked one at a time but read in batches. The buffer is
+	 * device major, so each device's chunks are contiguous.
+	 */
+	unsigned long long batchf = min(length, CHECK_BATCH_BYTES /
+				 ((unsigned long long)raid_disks * chunk_size));
+	size_t device_span = batchf * chunk_size;
+
+	if (posix_memalign((void**)&stripe_buf, 4096, raid_disks * device_span) != 0)
```
mtkaczyk: checkpatch is right here. it should be:

```
	if (posix_memalign((void **)&stripe_buf, 4096, raid_disks * device_span) != 0)
```

raid6check.c
```diff
@@ -389,110 +396,124 @@ int check_stripes(struct mdinfo *info, int *source, unsigned long long *offsets,
...
+			diskP = geo_map(-1, stripe, raid_disks, level, layout);
+			block_index_for_slot[-1] = diskP;
+			blocks[-1] = stripes[diskP];
+
+			diskQ = geo_map(-2, stripe, raid_disks, level, layout);
+			block_index_for_slot[-2] = diskQ;
+			blocks[-2] = stripes[diskQ];
+
+			if (!is_ddf(layout)) {
+				/* The syndrome-order of disks starts immediately after 'Q',
```
mtkaczyk, line 445: the comment has to be reworked to:
```
/*
 * The syndrome-order of disks starts immediately after 'Q', but skips `P`.
 */
```

or 
```
/* The syndrome-order of disks starts immediately after 'Q', but skips `P.` */
```

raid6check.c
```diff
@@ -389,110 +396,124 @@ int check_stripes(struct mdinfo *info, int *source, unsigned long long *offsets,
...
+					} else {
+						blocks[i] = stripes[i];
+						block_index_for_slot[i] = i;
+					}
 				}
 			}
-		}
 
-		qsyndrome(p, q, (uint8_t**)blocks, syndrome_disks, chunk_size);
+			qsyndrome(p, q, (uint8_t**)blocks, syndrome_disks, chunk_size);
```
mtkaczyk: ```qsyndrome(p, q, (uint8_t **)blocks, syndrome_disks, chunk_size);```

raid6check.c
```diff
@@ -389,110 +396,124 @@ int check_stripes(struct mdinfo *info, int *source, unsigned long long *offsets,
...
-				int slot = block_index_for_slot[role];
-				if (slot >= 0)
-					printf("Error detected at stripe %llu, page %d: possible failed disk slot %d: %d --> %s\n",
-					       start, j, role, slot, name[slot]);
-				else
-					printf("Error detected at stripe %llu, page %d: failed slot %d should be zeros\n",
-					       start, j, role);
-			} else if(disk[j] == -65535) {
-				printf("Error detected at stripe %llu, page %d: disk slot unknown\n", start, j);
+			for(j = 0; j < (chunk_size >> CHECK_PAGE_BITS); j++) {
```
mtkaczyk:
```suggestion
			for (j = 0; j < (chunk_size >> CHECK_PAGE_BITS); j++) {
```

raid6check.c
```diff
@@ -389,110 +396,124 @@ int check_stripes(struct mdinfo *info, int *source, unsigned long long *offsets,
...
+				int role = disk[j];
```
mtkaczyk, line 481: please add blank line

raid6check.c
```diff
@@ -389,110 +396,124 @@ int check_stripes(struct mdinfo *info, int *source, unsigned long long *offsets,
...
+				if (role >= -2) {
+					int slot = block_index_for_slot[role];
```
mtkaczyk, line 484: please add blank line

raid6check.c
```diff
@@ -389,110 +396,124 @@ int check_stripes(struct mdinfo *info, int *source, unsigned long long *offsets,
...
+					if (slot >= 0)
+						printf("Error detected at stripe %llu, page %d: possible failed disk slot %d: %d --> %s\n",
+						       stripe, j, role, slot, name[slot]);
+					else
+						printf("Error detected at stripe %llu, page %d: failed slot %d should be zeros\n",
+						       stripe, j, role);
+				} else if(disk[j] == -65535) {
```
mtkaczyk:
```suggestion
				} else if (disk[j] == -65535) {
```

mtkaczyk (commented):
Hey,
You moved a lot of code so please fix these small issues. It is not worthy keeping them.

For the  too long line issues, I'm fine with them. You don't have to rework messages.

I hope that is all.

@ncroxon are you executing these tests in your daily job? Do you want to comment?

ncroxon:
> @ncroxon are you executing these tests in your daily job? Do you want to comment?

In this particular case, I have not pulled in this patch and tried it.  
All depends on what I'm trying to accomplish in my day job.

davidgfnet:
I had this commit ready for a later PR, but here we go. This should fix style issues reported by the linter.
Thanks!

raid6check.c
```diff
@@ -357,24 +361,27 @@ int check_stripes(struct mdinfo *info, int *source, unsigned long long *offsets,
...
-	if (posix_memalign((void**)&stripe_buf, 4096, raid_disks * chunk_size) != 0)
+	/* Stripes are checked one at a time but read in batches. The buffer is
+	 * device major, so each device's chunks are contiguous.
+	 */
+	unsigned long long batchf = min(length, CHECK_BATCH_BYTES /
+				 ((unsigned long long)raid_disks * chunk_size));
+	size_t device_span = batchf * chunk_size;
+
+	if (posix_memalign((void **)&stripe_buf, 4096, raid_disks * device_span) != 0)
 		exit(4);
```
mwilck: While it's true that 512 MiB is not excessive these days, this change strongly increases the likelihood of a fatal error, and it's well possible that this code runs on a system with tight memory. Can we just fall back to the old, slow algorithm when the allocation fails?

Also, perhaps we should use `sysconf(_SC_PAGESIZE)` here instead of hardcoding 4k?

raid6check.c
```diff
@@ -389,110 +396,128 @@ int check_stripes(struct mdinfo *info, int *source, unsigned long long *offsets,
 				err = -1;
 				goto exitCheck;
 			}
-			int read_res = read(source[i], stripes[i], chunk_size);
-			if (read_res < chunk_size) {
+			ssize_t read_res = read(source[i], stripe_buf + i * device_span,
+						batch_bytes);
+			if (read_res < (ssize_t)batch_bytes) {
```
mwilck: We read a lot more bytes than before now. It's unlikely but not impossible that a short read may happen here. It would be safer to wrap the read() call in a loop.

Btw, have you considered simply using `mmap()` for the chunks intead of `read()`?

mwilck (commented):
Some comments from my side.

raid6check.c
davidgfnet: Good idea, done!

raid6check.c
davidgfnet: Using O_DIRECT and mmap... are kinda incompatible I guess?
I wasn't trying to rewrite the whole thing, therefore I used read() since it was already being used so.
Anyway I have another PR coming up that uses async I/O for even faster reads, so I'd rather not move to mmap I think.

davidgfnet:
Please check the latest changes and let me know :)

ncroxon:
@davidgfnet still a lot of failures from checkpatch.
https://github.com/md-raid-utilities/mdadm/actions/runs/32184414163/job/98766934333?pr=288

davidgfnet:
Sorry about that, added a commit to address most of that (ran checkpath.pl on the file).
It seems there are still some warnings like:

Error: WARNING:SPLIT_STRING: quoted string split across lines

And some others like:
WARNING: externs should be avoided in .c files
#47: FILE: raid6check.c:47:
+void qsyndrome(uint8_t *p, uint8_t *q, uint8_t **sources, int disks, int size);

Lemme know if (and how) I should handle these.
Thanks!

PR #290, opened 2026-08-11
super-ddf: fix mppe OOB read in examine_vd

ncroxon:
DDF record validation and the examine_vd() iteration bounds.

Validation of conf_rec_len / mppe in load_ddf_global() new helper ddf_max_primary_entries() to compute safe entry capacity bounds-limited examine_vd() loop using validated record size.
This prevents the phys_refnum[i] and
LBA_OFFSET(sb, vc)[i] OOB reads.

V2:
fix duplicate max_count initialization
make initialization reversed christmast tree
add a error message when failed to load the global infor.

### DIFF
--- a/super-ddf.c
+++ b/super-ddf.c
@@ -1133,7 +1133,23 @@ static int load_ddf_global(int fd, struct ddf_super *super, char *devname)
 	super->max_part = be16_to_cpu(super->active->max_partitions);
 	super->mppe = be16_to_cpu(super->active->max_primary_element_entries);
 	super->conf_rec_len = be16_to_cpu(super->active->config_record_len);
+	if (super->conf_rec_len * 512 <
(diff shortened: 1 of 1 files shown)


mtkaczyk (approved):

XiaoNi87 (commented):
Hi Nigel

load_super_ddf_all calls load_ddf_global. You add a new condition check in load_ddf_global which may returns error. It doesn't check the error return in load_super_ddf_all. So it still can face the error you fix. It needs to check the error return value in load_super_ddf_all.

I know it's a historical problem. But to avoid seeing this problem again from customer. It's better to add the check in load_super_ddf_all.

XiaoNi87 (approved):
This patch doesn't resolve the historical problem. But this patch looks good to me. Acked.

PR #291, opened 2026-08-11
[PATCH] add helper when calculating array superblock buffer

ncroxon:
A malformed MD v1 superblock can contain max_dev larger than the fixed 4096-byte superblock buffer. Before this fix, calc_sb_1_csum() and examine_super1() could read past the buffer and trigger out-of-bounds reads.

Added a helper sb_max_dev() that reads sb->max_dev from the on-disk v1 superblock and clamps it to MAX_DEVS. Updated calc_sb_1_csum() to compute checksum length using the clamped max_dev instead of the raw on-disk value. Updated examine_super1() to iterate dev_roles[] only up to the clamped max_dev when calculating array state and consistency.

The change prevents those unsafe reads by bounding max_dev to the maximum number of device role entries that fit in MAX_SB_SIZE.

### DIFF
--- a/super1.c
+++ b/super1.c
@@ -29,13 +29,29 @@
 #define MAX_SB_SIZE 4096
 /* bitmap super size is 256, but we round up to a sector for alignment */
 #define BM_SUPER_SIZE 512
-#define MAX_DEVS ((int)(MAX_SB_SIZE - sizeof(struct mdp_superblock_1)) / 2)
+#define DEV_ROLE_ALIGN 256
+#define MAX_DEVS ((int)((MAX_SB_SIZE - sizeof(struct mdp_superblock_1)) \
+			 / 2) & ~(DEV_ROLE_ALIGN - 1))
 #define SUPER1_SIZE	(MAX_SB_SIZE + BM_SUPER_SIZE \
 			 + sizeof(struct misc_dev_info))
 
 struct misc_dev_info {
 	__u64 device_size;
 };
+
+static unsigned int sb_max_dev(const struct mdp_superblock_1 *sb)
+{
+	unsigned int max_dev = __le32_to_cpu(sb->max_dev);
+
+	return max_dev > MAX_DEVS ? MAX_DEVS : max_dev;
+}
+
+static unsigned int sb_max_dev_aligned(unsigned int max_dev)
+{
+	if (max_dev > MAX_DEVS)
+		max_dev = MAX_DEVS;
+	return (max_dev + DEV_ROLE_ALIGN - 1) & ~(DEV_ROLE_ALIGN - 1);
+}
 
 #define MULTIPLE_PPL_AREA_SIZE_SUPER1 (1024 * 1024) /* Size of the whole
 						     * mutliple PPL area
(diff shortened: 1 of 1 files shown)


super1.c
```diff
@@ -37,6 +37,13 @@ struct misc_dev_info {
 	__u64 device_size;
 };
 
+static unsigned int sb_max_dev(const struct mdp_superblock_1 *sb)
+{
+	unsigned int max_dev = __le32_to_cpu(sb->max_dev);
```
mtkaczyk, line 44: max_devs should be also aligned to full sector. Please include this verification too.

XiaoNi87: if (max_dev > MAX_DEVS)
    max_dev = MAX_DEVS;

size = sizeof(*sb) + max_dev * 2;
size &= ~(SB_SECTOR_SIZE - 1);
if (size < sizeof(*sb))
    return 0;

return (size - sizeof(*sb)) / 2;

something like this?

ncroxon: The key point is that the stored role array is 2 bytes per device entry, so the alignment is based on 256 entries per 512-byte sector:
512 bytes / 2 bytes per entry = 256 entries

So the aligned maximum is computed in units of 256 entries, not by rounding the total byte size afterward.

The actual fix was to constrain and expand the device count at the entry-count level, using sector alignment in the role table boundary logic. In other words, the logic is centered on the max device count itself, not on computing a rounded byte size and then converting it back to a count.

Xiao, you snippet is close but different
Your version:
clamps the value, rounds the byte size to sector size, then converts back to an entry count.

That is conceptually similar, but it is not the same as the fix applied here. This patch keeps the alignment at the device-role table boundary and ensures each expansion step respects the sector-aligned maximum.

super1.c
```diff
@@ -29,14 +29,30 @@
 #define MAX_SB_SIZE 4096
 /* bitmap super size is 256, but we round up to a sector for alignment */
 #define BM_SUPER_SIZE 512
-#define MAX_DEVS ((int)(MAX_SB_SIZE - sizeof(struct mdp_superblock_1)) / 2)
+#define DEV_ROLE_ALIGN 256
+#define MAX_DEVS ((int)((MAX_SB_SIZE - sizeof(struct mdp_superblock_1)) \
+			 / 2) & ~(DEV_ROLE_ALIGN - 1))
```
mtkaczyk, line 34: This is not what I meant. I found somewhere info that when calculating max_dev first time we take the disks_count and we narrow it down to closest full sector:
for example:
MAX_SUPER_SIZE is 4096, SUPERBLOCK has always 256 and array has two members so we will set max_dev to 128 (superblock + devroles are 512 now so full sector is saved).

I failed to find reference now:( 

Given said, I think that allowed values for max_dev are 128 (one sector), 386 (two sectors), 642 (three sectors), 898 (four sectors)... up to 8 sectors.

This is definitely no about shrinking the max_devs accordingly to the expected alignment which may lead us to lose of some devs from the array. It is to late, we have to respected what it is, just ensure we are not crossing expected boundaries (4096) which may caused overlap with other reserved areas (bblog is next).

After more careful analysis I would like to revert my words about verifying against full sector alignment.
@ncroxon sorry for confusion.

ncroxon: I'm confused now...

mtkaczyk: clarified offine

mtkaczyk (changes requested):
Waiting for updated version

PR #292, opened 2026-08-17
fix IMSM zero-length allocation and out-of-bounds write

ncroxon:
The root cause is the anchor->mpb_size (a raw on-disk __u32) was used directly, unvalidated and even without endian conversion, to compute super->len and size the posix_memalign() allocation. A crafted/corrupt device with mpb_size == 0 makes super->len == 0, so posix_memalign returns a zero-length (or minimal) buffer, and the very next line unconditionally does memcpy(super->buf, anchor, sector_size) — writing a full sector into a buffer too small to hold it.

The fix is to read mpb_size with proper __le32_to_cpu() conversion Everything downstream (super->len = ROUND_UP(mpb_size, sector_size), the allocation, and the memcpy) now uses this validated, endian-corrected mpb_size, so a zero-length allocation followed by a full-sector memcpy can no longer happen.

### DIFF
--- a/super-intel.c
+++ b/super-intel.c
@@ -4627,12 +4627,29 @@ static int load_imsm_mpb(int fd, struct intel_super *super, char *devname)
 		return 2;
 	}
 
+	/* Sanity check the on-disk mpb_size before trusting it for the
+	 * allocation below. A value of 0 (or anything smaller than the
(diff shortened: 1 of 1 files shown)


mtkaczyk (approved):
LGTM

XiaoNi87 (approved):
Thanks for the patch.

PR #293, opened 2026-08-19
super-ddf: clamp max_pd_entries in load_ddf_local to prevent OOB read

cenhuil:
max_pd_entries is taken from an untrusted DDF header without clamping it to the actual size of the phys_section. This could lead to a heap out-of-bounds read when iterating over phys_disk entries.

Use ddf_safe_entry_count() to clamp the loop boundary.

### DIFF
--- a/super-ddf.c
+++ b/super-ddf.c
@@ -1195,6 +1195,7 @@ static int load_ddf_local(int fd, struct ddf_super *super,
 	int vnum;
 	unsigned int max_virt_disks =
 		be16_to_cpu(super->active->max_vd_entries);
+	unsigned int max_pdes;
 	unsigned long long dsize;
 
 	/* First the local disk info */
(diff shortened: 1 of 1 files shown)


mwilck (approved):
LGTM

super-ddf.c
```diff
@@ -1237,10 +1237,17 @@ static int load_ddf_local(int fd, struct ddf_super *super,
 		dl->vlist[i] = NULL;
 	super->dlist = dl;
 	dl->pdnum = -1;
-	for (i = 0; i < be16_to_cpu(super->active->max_pd_entries); i++)
-		if (memcmp(super->phys->entries[i].guid,
-			   dl->disk.guid, DDF_GUID_LEN) == 0)
-			dl->pdnum = i;
+	{
+		unsigned int max_pdes = ddf_safe_entry_count(super->pdsize,
```
mtkaczyk: do not cheat on checkpatch please :)

This is not a new function so {} are not necessary

mtkaczyk (changes requested):

super-ddf.c
mwilck: I think he wanted to ensure "local" scope of `max_pdes`. But I didn't like this style, either.

cenhuil: I have modified the code. please review it.

mtkaczyk (approved):

PR #295, opened 2026-08-20
Examine: Add rsync offset for native

mkusiak127:
During rebuild, mdadm examine will display "Recovery Offset" entry if recovery is active. Add displaying "Resync Offset" for 0.90 and 1.x so the behavior is consistent. Usefull for checking resync progress on offline array.

### DIFF
--- a/super0.c
+++ b/super0.c
@@ -186,6 +186,10 @@ static void examine_super0(struct supertype *st, char *homehost)
 	printf("    Update Time : %.24s\n", ctime(&atime));
 	printf("          State : %s\n",
 	       (sb->state&(1 << MD_SB_CLEAN)) ? "clean":"active");
+	/* recovery_cp is a legacy name, from before sync actions got split */
+	if (sb->recovery_cp != (__u32)MaxSector)
+		printf("  Resync Offset : %llu sectors\n",
+		       (unsigned long long)sb->recovery_cp);
 	if (sb->state & (1 << MD_SB_BITMAP_PRESENT))
 		printf("Internal Bitmap : present\n");
 	printf(" Active Devices : %d\n", sb->active_disks);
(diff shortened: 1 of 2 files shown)


super0.c
```diff
@@ -186,6 +186,9 @@ static void examine_super0(struct supertype *st, char *homehost)
 	printf("    Update Time : %.24s\n", ctime(&atime));
 	printf("          State : %s\n",
 	       (sb->state&(1 << MD_SB_CLEAN)) ? "clean":"active");
+	if (sb->recovery_cp + 1)
```
mtkaczyk: This is recovery offset but you are printing it as resync checkpoint, are you sure?

super1.c
```diff
@@ -298,6 +298,9 @@ static void examine_super1(struct supertype *st, char *homehost)
 	if (__le32_to_cpu(sb->feature_map) & MD_FEATURE_RECOVERY_OFFSET)
 		printf("Recovery Offset : %llu sectors\n",
 		       (unsigned long long)__le64_to_cpu(sb->recovery_offset));
+	if (__le64_to_cpu(sb->resync_offset) + 1)
```
mtkaczyk: I don't like that, but fine.

I would preffer to directly compare with invalid_resync_offset

super0.c
mkusiak127: Added a note this is a (pre split) legacy name + compared to `MaxSector`: [changes](https://github.com/md-raid-utilities/mdadm/compare/4dbcd6f4f45d9a44613123e805a8ba41209c0309..f46f8c07a8164f0e9f6950d616d63347067992f1)

You can verify here it is truly handling resync https://github.com/md-raid-utilities/mdadm/blob/406ab49fa94b8ba5eda88d82673d9d2b1692e185/super0.c#L667

super1.c
mkusiak127: [Compared to MaxSector](https://github.com/md-raid-utilities/mdadm/compare/4dbcd6f4f45d9a44613123e805a8ba41209c0309..f46f8c07a8164f0e9f6950d616d63347067992f1)

XiaoNi87 (approved):
Looks good to me.

mtkaczyk (approved):

PR #296, opened 2026-08-27
super1: validate max_dev heap

ncroxon:
Fix this out-of-bounds read vulnerability by validating sb->max_dev immediately after the superblock is loaded in load_super1().

The fix involves checking __le32_to_cpu(super->max_dev) > MAX_DEVS (where MAX_DEVS is (4096 - 256) / 2 = 1920). Rejecting corrupt superblocks at load time prevents out-of-bounds reads across calc_sb_1_csum(), examine_super1(), and getinfo_super1().

### DIFF
--- a/super1.c
+++ b/super1.c
@@ -2207,6 +2207,14 @@ static int load_super1(struct supertype *st, int fd, char *devname)
 		return 1;
 	}
 
+	if (__le32_to_cpu(super->max_dev) > MAX_DEVS) {
+		if (devname)
+		    pr_err("max_dev (%u) exceeds maximum allowed (%d) in superblock on %s\n",
+			__le32_to_cpu(super->max_dev), MAX_DEVS, devname);
+		free(super);
+		return 2;
+	}
+
 	if (__le32_to_cpu(super->magic) != MD_SB_MAGIC) {
 		if (devname)
 			pr_err("No super block found on %s (Expected magic %08x, got %08x)\n",


ncroxon:
This patch is for https://github.com/md-raid-utilities/mdadm/security/advisories/GHSA-5h5p-mx5j-6q4j

super1.c
```diff
@@ -2210,6 +2207,14 @@ static int load_super1(struct supertype *st, int fd, char *devname)
 		return 1;
 	}
 
+	if (__le32_to_cpu(super->max_dev) > MAX_DEVS) {
+		if (devname)
+		    pr_err("max_dev (%u) exceeds maximum allowed (%d) in superblock on %s\n",
+			__le32_to_cpu(super->max_dev), MAX_DEVS, devname);
```
mtkaczyk, line 2213: Bad code indent here:
```diff 
-			__le32_to_cpu(super->max_dev), MAX_DEVS, devname);
+				 __le32_to_cpu(super->max_dev), MAX_DEVS, devname);
```

mtkaczyk: Also, please rebase this if you can to have clean one commit.

mtkaczyk (changes requested):

super1.c
```diff
@@ -2210,6 +2210,14 @@ static int load_super1(struct supertype *st, int fd, char *devname)
 		return 1;
 	}
 
+	if (__le32_to_cpu(super->max_dev) > MAX_DEVS) {
+		if (devname)
```
mwilck, line 2214: If we need to print a message here, I suggest using something like `(devname ? devname : "*unknown device*") here rather than skipping the message entirely when `devname` is `NULL`.

PR #297, opened 2026-08-31
.gitignore: Add sandbox

mkusiak127:
Add generic `sandbox` entry to gitignore. Handy for storing logs, scripts, etc. within the codebase context, untracked.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,6 @@
 /mdassemble.static
 /mdmon.O2
 /raid6check
+
+# Generic untracked directory for logs, scripts etc. for local development.
+sandbox/


mtkaczyk (approved):
I'm fine with that. Won't hurt.

mwilck:
Are you aware that there's `core.excludesFile` in `~/.gitconfig` ?

To me this looks like personal preference of @mkusiak127 rather than something that we should enforce for everyone.

mkusiak127:
> Are you aware that there's `core.excludesFile` in `~/.gitconfig` ?

Nope, I was not.

> To me this looks like personal preference of @mkusiak127 rather than something that we should enforce for everyone.

I picked this up from other projects I worked on, though others might get use out of it.  I don't insist on it though, feel free to close this PR if you don't like it.

mwilck:
`.gitignore` has a couple of surprising side effects. For example, **ripgrep** skips ignored files or directories by default.

While I see the benefit that this has for individual workflows, I'd prefer limiting the project's `.gitignore` to files that are created by the build process.

PR #299, opened 2026-09-02
mdadm: Set appropriate bitmap state during grow

bjohnsto:
Add check in the grow code to make sure we set the bitmap major value to lockless so it properly fails and does not create an internal bitmap in its place. Add a test to prove this fix works.

Add error and doc fixes for lockless bitmap support as well.


Grow.c
```diff
@@ -298,6 +298,9 @@ int Grow_addbitmap(char *devname, int fd, struct context *c, struct shape *s)
 	if (s->btype == BitmapCluster)
 		major = BITMAP_MAJOR_CLUSTERED;
 
+	if (s->btype == BitmapLockless)
```
mwilck: nitpick: using `else if` would improve readability here.

mwilck (commented):
LGTM, just one nit.

Grow.c
bjohnsto: Good point. I'll make this fix

tests/05r1-grow-lockless
```diff
@@ -0,0 +1,16 @@
+#
+# lockless bitmaps are only supported at array-creation time with 1.2
+# metadata; verify --grow --bitmap=lockless is rejected and does not
+# silently add a regular internal bitmap.
+#
+mdadm --create --run $md0 --metadata=1.2 --level=1 -n2 --delay=1 $dev1 $dev2
```
mtkaczyk: please specify directly that bitmap should be none `--bitmap=none`. This is skipped with `--run` but you must know that to make it direct please :)

Grow.c
bjohnsto: I updated the Grow code to use else if

tests/05r1-grow-lockless
bjohnsto: updated

mtkaczyk (approved):

mtkaczyk (approved):

PR #300, opened 2026-09-07
Fixtests

XiaoNi87:
This patch set is used to fix regression test failure.


tests/07autodetect
```diff
@@ -5,7 +5,9 @@
 # put a partition on each, create an array, and see if we can
 # use autodetect to restart the array.
 
-[if lsmod | grep md_mod > /dev/null 2>&1] || skip "md is a module - cannot test autodetect"
+if lsmod | grep -q '^md_mod[[:space:]]'; then
```
mwilck: `grep -q '^md_mod\>' /proc/modules` ?

Grow.c
```diff
@@ -4306,6 +4306,15 @@ static int grow_backup(struct mdinfo *sra,
 
 	if (rv)
 		return rv;
+
+	/*
+	 * Make backup data durable before publishing metadata that marks it
+	 * usable after a crash.
+	 */
+	for (i = 0; i < dests; i++)
+		if (fsync(destfd[i]) != 0)
```
mwilck: Should an error message be emitted here?

Grow.c
```diff
@@ -4330,7 +4339,8 @@ static int grow_backup(struct mdinfo *sra,
 			if (write(destfd[i], &bsb, 512) != 512)
 				break;
 		}
-		fsync(destfd[i]);
+		if (fsync(destfd[i]) != 0)
```
mwilck: Should an error message be emitted here?

Makefile
```diff
@@ -114,6 +114,8 @@ CHECK_RUN_DIR=1
 MAP_DIR=$(RUN_DIR)
 MAP_FILE = map
 MAP_PATH = $(MAP_DIR)/$(MAP_FILE)
+# Reshape backup file links must survive a reboot.
+BACKUP_DIR=/var/lib/mdadm
```
mwilck: Add whitespace around `=`

Grow.c
```diff
@@ -1171,18 +1171,25 @@ int reshape_open_backup_file(char *backup_file,
...
+			free(bu);
+			return 1;
+		}
 		if (backup_file[0] != '/') {
 			resolved = realpath(backup_file, NULL);
 			if (resolved)
 				target = resolved;
 		}
 		if (symlink(target, bu))
-			pr_err("Recording backup file in " MAP_DIR " failed: %s\n",
```
mwilck, line 1180: `backup_file` is now persistent. AFAICS, `reshape_array()` unlinks the file only if succesful (`if (backup_file && done)`). Can it happen that a stale backup file survives there after a failed or aborted reshape? If yes, what would be the consequences?

tests/07autoassemble
```diff
@@ -2,6 +2,15 @@
...
 # tear it down and get auto-assemble to rebuild it.
 
+active_raid0_dev()
+{
+	local dev
+
+	dev=$(awk '$3 == "active" && $4 == "raid0" { print "/dev/" $1; exit }' /proc/mdstat)
+	[ -n "$dev" ] || die "active raid0 not found"
+	echo "$dev"
+}
```
mwilck, line 14: While this is certainly more robust than before, if the test system has any RAID0 arrays before starting the test, it can still return the wrong one. Or am I missing something?

tests/07autodetect
mwilck: `grep -q '^md_mod\>' /proc/modules` ?

tests/07reshape5intr
```diff
@@ -2,10 +2,74 @@
 #
 # test interrupting and restarting raid5 reshape.
 set -x
+
+bu=$targetdir/md-backup
```
mwilck: I know you're just writing this code in the style that these scripts have been written in before, but I don't like this sort of global variables in shell scripts. I would at least recommend to make globals recognizable by using `CAPITALS` for their names to clearly distinguish them from locals.

tests/07reshape5intr
```diff
@@ -2,10 +2,74 @@
...
+saved_bu=$targetdir/md-backup.saved
+map_bu=/run/mdadm/backup_file-md0
+rm -f $bu $saved_bu $map_bu
+
+wait_for_reshape_start() {
+  cnt=100
```
mwilck: Likewise, I'd suggest using `local cnt=100` here to make sure to avoid any name clashes of this rather generically named variable.

tests/07reshape5intr
```diff
@@ -22,34 +86,27 @@ do
   mdadm $md0 --add $dev6
   echo 20 > /proc/sys/dev/raid/speed_limit_min
   echo 20 > /proc/sys/dev/raid/speed_limit_max
-  mdadm --grow $md0 -n $[disks+1]
+  mdadm --grow $md0 -n $[disks+1] --backup-file=$bu
+  need_backup=0
+  grep -q "Need to backup" $targetdir/stderr && need_backup=1
+  wait_for_reshape_start
+  preserve_backup || [ $need_backup -eq 0 ] ||
```
mwilck: Why is it necessary to make a copy of the backup file? The commit messsage doesn't explain it. Can it be overwritten by some other process? If yes, how can we be certaint that backup of the backup can't be overwritten as well?

tests/func.sh
```diff
@@ -213,13 +213,21 @@ restore_selinux() {
 wait_for_reshape_end() {
 	# wait for grow-continue to finish but break if sync_action does not
 	# contain any reshape value
+	local grow_continue_timeout=30
```
mwilck: For time-based wait loops like this, I generally prefer something like
```
local deadline=$(("$(date +%s)" + 30))
while [[ "$(date +%s)" -lt "$deadline" ]]; do
   ...
   sleep 1
done
```

tests/09imsm-assemble
```diff
@@ -20,17 +20,29 @@ imsm_check_removal() {
...
-        echo "$2 removal from $1 should have succeeded" >&2
+        echo "$2 should have fallen out of $1 or been removable" >&2
         cat /proc/mdstat >&2
         mdadm -E "$2"
         exit 1
     fi
 }
 
+imsm_member_has_disk() {
+    mdadm -D "$1" | awk -v dev="$2" '$NF == dev { found=1 } END { exit !found }'
```
mwilck: Do you need to check for `active` state here?

Also, would it be good to be sure to skip everything before the RAID device list? E.g. like this:
```
awk -v dev="$2" '/RaidDevice/ { next; } prnt == 1 && $NF == dev { found=1; } END { exit !found }'
```

tests/09imsm-create-fail-rebuild
```diff
@@ -28,7 +28,7 @@ mdadm -CR $container -e imsm -n $num_disks $dev0 $dev1
 imsm_check container $num_disks
 
 # RAID0 + RAID1
-size=9000
+size=7000
```
mwilck, line 31: Could you explain why this change is necessary?

tests/10ddf-incremental-wrong-order
```diff
@@ -41,7 +41,7 @@ sleep 2
 seq8a=$(mdadm -E $dev8 | sed -n 's/^ *Seq : //p')
 seq10a=$(mdadm -E $dev10 | sed -n 's/^ *Seq : //p')
 
-if [ $seq8a -ge $seq10a ]; then
+if [ $((16#$seq8a)) -ge $((16#$seq10a)) ]; then
```
mwilck: I'd prefer `0x$seq8a` here, as the `16#$seq8a` syntax, while  correct, is not widely known (at least I wasn't familiar with it 😄 )

tests/10ddf-incremental-wrong-order
```diff
@@ -52,35 +52,35 @@ fi
...
 
-seq8c=$(mdadm -E $dev8 | sed -n 's/^ *Seq : //p')
-seq10c=$(mdadm -E $dev10 | sed -n 's/^ *Seq : //p')
-
 if [ x$sha_0a != x$sha_0c ]; then
     ret=1
     echo ERROR: sha1sum of $member0 has changed
 fi
-if [ x$sha_1b != x$sha_1c ]; then
+if [ -n "$sha_1c" ] && [ x$sha_1b != x$sha_1c ]; then
```
mwilck: you can simply write `if [[ $sha_1b != $sha_1c ]]` here.

tests/10ddf-incremental-wrong-order
```diff
@@ -114,7 +114,8 @@ if mdadm -A $container $dev8 $dev9 $dev10 $dev11; then
 	ret=1
 	echo ERROR: sha1sum of $member1 has changed
     fi
-    if [ \( $seq10a -ge $seq10d \) -o \( $seq8d -ne $seq10d \) ]; then
+    if [ $((16#$seq10a)) -ge $((16#$seq10d)) ] ||
```
mwilck: I suggest using bash syntax here for better robustness and readability.
```
if [[ $((0x$seq8a)) > $((0x$seq10d)) || $seq8d != $seq10d ]]; then
```
Unfortunately, the hex->dec conversion remains a bit messy.

tests/19raid6auto-repair
```diff
@@ -26,6 +23,8 @@ do
     echo 3 > /proc/sys/vm/drop_caches
     cmp -s -n $array_data_size_in_b $md0 /tmp/RandFile || { echo sanity cmp failed ; exit 2; }
 
+    data_offset_in_kib=$[$(cat /sys/block/${md0##*/}/md/rd0/offset)/2]
```
mwilck: Please use `$(())` instead of use the legacy `$[]` syntax.

tests/19raid6repair
```diff
@@ -22,11 +19,12 @@ do
 	    device_with_error=${failure_split[0]}
 	    stripe_with_error=${failure_split[1]}
 	    repair_params="$stripe_with_error ${failure_split[2]} ${failure_split[3]}"
-	    start_of_errors_in_kib=$[data_offset_in_kib+chunksize_in_kib*stripe_with_error]
 
 	    # make a raid5 from a file
 	    dd if=/dev/urandom of=/tmp/RandFile bs=1024 count=$array_data_size_in_kib
 	    mdadm -CR $md0 -l6 --layout=$layout -n$number_of_disks -c $chunksize_in_kib $devs
+	    data_offset_in_kib=$[$(cat /sys/block/${md0##*/}/md/rd0/offset)/2]
+	    start_of_errors_in_kib=$[data_offset_in_kib+chunksize_in_kib*stripe_with_error]
```
mwilck, line 27: Please use `$(())`.

test
```diff
@@ -383,7 +395,12 @@ main() {
 		then
 			for script in ${TESTLIST[@]}
 			do
-				do_test $testdir/$script
+				if ! script_path=$(resolve_test_path "$script"); then
+					echo "Invalid script: '$script'." >&2
```
mwilck: Move this error message into `resolve_test_path` ?

Grow.c
```diff
@@ -910,13 +910,16 @@ static int subarray_set_num(char *container, struct mdinfo *sra, char *name, int
...
-	err = sysfs_set_num(sra, NULL, "suspend_hi", sra->reshape_progress);
-	err = err ?: sysfs_set_num(sra, NULL, "suspend_lo",
-				   sra->reshape_progress);
+	if (!already_running) {
+		sysfs_set_num(sra, NULL, "suspend_lo", 0x7FFFFFFFFFFFFFFFULL);
+		err = sysfs_set_num(sra, NULL, "suspend_hi",
+				    sra->reshape_progress);
+		err = err ?: sysfs_set_num(sra, NULL, "suspend_lo",
+					   sra->reshape_progress);
+	}
```
mwilck, line 922: Is this commit related to [CVE-2026-53123](https://www.cve.org/CVERecord?id=CVE-2026-53123)?

Grow.c
```diff
@@ -1179,21 +1179,32 @@ int reshape_open_backup_file(char *backup_file,
...
 		char *resolved = NULL;
+		char *tmp;
+		int err;
 
 		if (mkdir(BACKUP_DIR, 0755) && errno != EEXIST) {
 			pr_err("Creating backup directory " BACKUP_DIR
 			       " failed: %s\n", strerror(errno));
 			free(bu);
-			return 1;
+			return 0;
```
mwilck: Why return success here? I fail to match this hunk with the commit description.

mwilck (commented):
Thanks for this very impressive patch set!

I have added some comments. I am not familiar with array reshape in detail, so my comments about that code may be mislead. Perhaps just some more explanations for these complex matters would be in order.

Grow.c
mwilck: I realize that you changed this code again in 05d1770, but I am not sure if that fix addresses the problem I mentioned here.

tests/07autoassemble
XiaoNi87: Yes, I'll fix in next version. 
Let's save the UUID of the array created by the test before stopping it. After reassembly, enumerate active RAID0 devices and accept only the device whose UUID, obtained with `mdadm --detail`, matches the saved UUID.

tests/07reshape5intr
XiaoNi87: I agree with you. This part should be deleted. I'll fix it in next version

tests/func.sh
XiaoNi87: sure, no problem

Grow.c
XiaoNi87: fix this in next version

Grow.c
XiaoNi87: sure

Makefile
XiaoNi87: fix it in next version

Grow.c
XiaoNi87: Thanks for pointing this out. I'll fix this in next vesion.

A backup file can indeed survive a failed or aborted reshape in the original implementation, because reshape_array() only removed backup_file when done was true.

The cleanup must distinguish between a reshape that never started and one that was already active. If backup preparation or impose_reshape()/start_reshape() fails before the reshape starts, the newly created backup file and its persistent locator are removed. If the reshape has already started and the monitor aborts, the locator is retained for now: abort_reshape() stops the kernel reshape but does not necessarily clear the reshape metadata. The backup may still be required to restore the critical section after a reboot or re-assembly. It is removed only after successful completion, or once the array metadata confirms that no active reshape remains.

The follow-up change also removes an old locator when the array is known not to have an active reshape. This removes stale locator state without following the symlink or deleting an unrelated target file.

tests/09imsm-assemble
XiaoNi87: Thanks for the suggestions. I've fixed the two problems in new version.

tests/09imsm-create-fail-rebuild
XiaoNi87: I've updated the new commit message in new version:

The test creates two IMSM volumes in each container.  With the 9000 KiB first
volume, IMSM alignment and reserved metadata space make the second volume
larger than the space available on the small loop devices used by the test
harness.  Array creation therefore reports a size mismatch before the fail and
rebuild scenario is reached.

Reduce the first volume to 7000 KiB so both aligned volumes fit and the test
can exercise the intended failure and rebuild path.  Remove the stale broken
marker.

tests/10ddf-incremental-wrong-order
XiaoNi87: This has been fixed in next version

tests/10ddf-incremental-wrong-order
XiaoNi87: Thanks. I changed the string comparison to use [[ ... ]], but kept the -n $sha_1c condition intentionally. This test allows member1 to remain unstarted during incremental assembly, so sha_1c is only assigned when the member device exists. Without the -n check, an unstarted member would compare its original checksum against an empty string and be incorrectly reported as data corruption. When member1 is started, its checksum is still validated normally.

tests/10ddf-incremental-wrong-order
XiaoNi87: ok, no problem.

tests/19raid6auto-repair
XiaoNi87: It has been fixed in next version

tests/19raid6repair
XiaoNi87: done

test
XiaoNi87: done

Grow.c
XiaoNi87: It has relationships with this cve, but it fixes userspace problem which can cause this deadlock

Grow.c
XiaoNi87: reshape_open_backup_file uses 0 as false return value.

tests/10ddf-incremental-wrong-order
mwilck: Thanks for the explanation.

tests/09imsm-assemble
XiaoNi87: We can't check active here. tests/09imsm-assemble will fail. Because the disks have following status:
  /dev/loop0  active sync
  /dev/loop1  spare rebuilding

dev1 is rejected even though it belongs to the selected family and is actively rebuilding. Neither family then matches, and the test reports:

  could not determine assembled IMSM family

mwilck (approved):
I went over this set in detail earlier this week. I just took another quick glance over it, as I currently lack time for another full review. 

That said, LGTM.

mtkaczyk (approved):

PR #301, opened 2026-09-08
tests: add regression test for raid 5 to 1 hang fix (RHEL-191802)

bjohnsto:
Add regression test for raid1_takeover() leaving conf->array_frozen set permanently after a 2-disk raid5->raid1 takeover, stalling all later I/O.

The test will be added in auto skip mode and will not be turned on until the kernel fix lands upstream and I have a way to gate on kernel version so the test doesn't hang on old kernels.


tests/07raid5-to-raid1-takeover-nofreeze
```diff
@@ -0,0 +1,87 @@
+# RHEL-191802: a 2-disk array taken over raid1 -> raid5 -> raid1 must not
+# come out permanently frozen -- otherwise every later I/O (including
+# umount) stalls forever in raid1's _wait_barrier(). Reproduces the
+# reported scenario directly: I/O between/after each takeover, then
+# unmount. Only the raid5->raid1 leg (raid1_takeover()) is actually
+# affected -- raid1->raid5 is exercised first purely to get there.
+
+# TODO: Replace with a kernel-version gate once the raid1_takeover()
+# array_frozen fix lands upstream
+skip "kernel fix for RHEL-191802 is not upstream yet"
```
mtkaczyk, line 10: we are not interested in Redhat defect Id in upstream, it is nice to have but it is not needed in skip message.

mtkaczyk: and you can add this when it is finally enabled in upstream, what is the purpose of tests that is always skipped?

tests/07raid5-to-raid1-takeover-nofreeze
```diff
@@ -0,0 +1,87 @@
+# RHEL-191802: a 2-disk array taken over raid1 -> raid5 -> raid1 must not
+# come out permanently frozen -- otherwise every later I/O (including
+# umount) stalls forever in raid1's _wait_barrier(). Reproduces the
```
mtkaczyk, line 3: At this point on time, I think we should say "stop" for introducing md regression testing in mdadm.

mdadm should handle tests that are used to verify mdadm logic i.e We make a mdadm change and we are running these tests to see if there is no regression in mdadm and these tests should be enabled in upstream mdadm repository.

Eventually, it should go to separate bucket of tests in mdadm that we know- it is md-regression only.

@mwilck @ncroxon I think that it is right moment to introduce repository that can be used for that in our organization, what do you think?

Of course, due to years of integration it won't be simple to isolate mdadm from md just for testing but I think it is time to start considering that.

I'm open for comments here.

mtkaczyk (commented):
Hey @bjohnsto 

Thanks for patch!
I added some comments, mostly challenging current state of mdadm tests so sorry for using your PR for further discussion :)

tests/07raid5-to-raid1-takeover-nofreeze
bjohnsto: I'm open to moving this or figuring out a better pattern. I do acknowledge the test is mostly created to test a kernel fix. On the other hand, testing takeover from m -> n -> m  feels like a valid thing for mdadm to test.

tests/07raid5-to-raid1-takeover-nofreeze
ncroxon: No reference to RHEL should be present in the headers.

tests/07raid5-to-raid1-takeover-nofreeze
ncroxon: @mtkaczyk, are you saying, create a repo for distro specific issues?

mtkaczyk: If distro specific regression affected upstream at any moment of time  (which should be usually the truth) why not? Usually all md patches are upstreamed.

You can even have your private repo that only mdadm-team can access.  We are working on this soft together so anything that make conversations and maintenance simpler is nice.

Please think and let me know what would be helpful.

mtkaczyk: >On the other hand, testing takeover from m -> n -> m feels like a valid thing for mdadm to test.

Indeed but in this particular test we are looking for regression.
If this test will not be skipped, then we would need to consider the value of having it here.

bjohnsto: > > On the other hand, testing takeover from m -> n -> m feels like a valid thing for mdadm to test.
> 
> Indeed but in this particular test we are looking for regression. If this test will not be skipped, then we would need to consider the value of having it here.

I would love to not skip but running on a non fixed kernel would mean the kernel would hang, which is why I'm skipping for now.

tests/07raid5-to-raid1-takeover-nofreeze
bjohnsto: yeah my bad. i'll clean that up

tests/07raid5-to-raid1-takeover-nofreeze
mtkaczyk: @mwilck please bring some brilliant idea how to handle this case :)

Adding "skipped" tests seems to be pointless, not adding it would make it never added, enabling make broke test suite because it would discover hung task.

I think that we should start grouping tests like that in a separate bucket like `md-regression tests`, and try filter mdadm focues, lower risk tests and advanced md-regression scenarios.

Of course, is something is fixed long time ago it can be still promoted to low-risk test, so maybe kind of tagging should be added to test suite?

mdadm and md are mixed and cannot be easily separated in current design. Probably that will never change..

bjohnsto: @mtkaczyk I am going to close this PR and move the test to somewhere else.

PR #303, opened 2026-09-16
Fix cTiB overflow

mwilck:
Suggested fix for #302

PR #305, opened 2026-09-18
Fix safeguard against deleting the last disk in a mirror

mwilck:
(This goes on top of #303, which it currently includes)

fc6fd4063769 ("Manage: Block unsafe member failing") was meant as a safeguard against users shooting themselves into the foot by removing the last up-to-date leg of a RAID mirror. But it works only if the RAID members are disks, not for partitions, and fails for `set-A` / `set-B` operations, too.

Add a test for this feature (which fails in the first place), and a fix that makes the test succeed.


mwilck:
Modified; based directly on main now, not on #303.

tests/01r1faillast
```diff
@@ -0,0 +1,39 @@
+# create a RAID1 from partitions, try to fail both legs.
```
mtkaczyk, line 1: Kernel `failfast` feature is something else. We have name conflict I think.

tests/01r1faillast
```diff
@@ -0,0 +1,39 @@
...
+# mdadm should refuse to fail the second one.
+
+if [[ $DEVTYPE = disk ]]; then
+    SEP=
+else
+    SEP=p
+fi
+
+mdadm -CR $md0 -l1 -n2 --assume-clean $dev2 $dev3
```
mtkaczyk, line 10: R1 and R10 covered, what about r5 or any from raid456?

tests/01r10faillast
```diff
@@ -0,0 +1,53 @@
...
+mdadm $md0 --fail $dev5
+[[ $? = 1 ]] || die "mdadm should return 1"
+check state _UU_
+
+read _st < /sys/block/md0/md/array_state
+[[ "$_st" = clean ]] || die "array state of $md0 is \"$_st\", should be \"clean\""
+mdadm -S "$md0"
+
+for d in $dev0 $dev1 $dev2 $dev3; do
+    sfdisk $d <<EOF
```
mtkaczyk, line 24: two cases are covered here:
- bug for partitions
- mdadm owned faulty state verification

Codex told me:
```
That exact partition-based example appears in [test line 195. Each expanded --disks= argument is 
collected into the device array at test line 346) then assigned to $dev0 through $dev13 in
tests/func.sh line 262

I found exactly one test under tests/ that creates partitions: [tests/07autodetect line 3.
It creates two partitionable MD arrays, runs sfdisk on those MD devices, and then uses ${mdp0}p1 and ${mdp1}p1.
It does not repartition the physical devices passed through --disks; the partition table is created inside
the MD arrays. No other test script uses sfdisk, fdisk, parted, partprobe, or kpartx.
```

I would prefer to not include partitions directly assuming that this defect will not appear in this form anymore and we should focus on executing this whole test suite on partitions and block devices as a two separate jobs.

Executing on partitions is very important case for MD stacked device. I know we have no working CI yet, but this should be our direction.

If I missed something, please let me know.

Manage.c
```diff
@@ -1327,15 +1327,15 @@ int Manage_with(struct supertype *tst, int fd, struct mddev_dev *dv,
  *
  * Return: True if array will be operational, false otherwise.
  */
-bool is_remove_safe(mdu_array_info_t *array, const int fd, char *devname, const int verbose)
+static bool is_remove_safe(mdu_array_info_t *array, const int fd, dev_t devid, const int verbose)
```
mtkaczyk, line 1330: description is not updated. When, updating please include why correct devid is important to be used here. You can mention the bug you discovered.

